PR #19, opened 2015-03-15
Update README.md with better install instructions

agundy:


seveibar:
can't be automatically merged- could you resolve this?

PR #23, opened 2015-03-15
User profile

agundy:
Small fixes to enable users to click to see a profile and makes sure a user has a github username when they create an account.


seveibar:
Looks good!

PR #34, opened 2015-03-18
Refactor the list and API so that correct pictures and relevant data are returned

kmcnellis:


agundy:
Can you fix it so that it passes Travis CI?

Issue #30, opened 2015-03-17
Project Blogs


agundy:
Projects should have blogs associated with them.

seveibar:
Is this fixed?

agundy:
yes

PR #41, opened 2015-03-20
Project Blogs

jfucci:
Added blogs for individual projects. Posts can be added but not edited or deleted at this point. The revert is only due to my starting on a way to add users to a project before realizing that it was different from how the ui is now implemented. Also, addresses #30 .

### DIFF
--- a/server/api/post/post.model.js
+++ b/server/api/post/post.model.js
@@ -1,0 +1,17 @@
+'use strict';
+
+var mongoose = require('mongoose'),
+    Schema = mongoose.Schema;
+
+var PostSchema = new Schema({
+  title: String,
+  content: String,
+  project: String,
+  author: {
+    name: String,
+    id: Object
+  },
+  date: Date
+});
+
+module.exports = mongoose.model('Post', PostSchema);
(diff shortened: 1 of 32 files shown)


agundy:
Can you fix it so that is passes Travis CI? Thanks!

server/api/post/post.model.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+var mongoose = require('mongoose'),
+    Schema = mongoose.Schema;
+
+var PostSchema = new Schema({
+  title: String,
+  content: String,
+  project: String,
```
seveibar: What if the project name changes? Should this be an id?

seveibar:
Awesome PR (at least w/ code stylistically). I'll pull it in and check functionality later (unless someone else gets to it haha)

server/api/post/post.model.js
jfucci: Good point. Changed to use ids.

seveibar:
I'm not sure this is ready for merge- the links on the projects page don't seem to go to a blog.

PR #44, opened 2015-03-28
Add commits to the user profile page as a separate call to the commits api

kmcnellis:


client/app/profile/profile.controller.js
```diff
@@ -4,5 +4,10 @@ angular.module('observatory3App')
   .controller('ProfileCtrl', function ($scope, $stateParams, $http) {
       $http.get('/api/users/' + $stateParams.id).success(function(user){
           $scope.user = user;
+          $http.get('/api/commits/user/' + user.githubProfile).success(function(commits){
+              $scope.user.commits = commits;
+              console.log(commits);
```
seveibar: remove log code

server/api/user/user.controller.js
```diff
@@ -57,18 +80,13 @@ exports.allStats = function(req, res) {
             .where('author.login').equals(String(user.github.login))
             .where('date').gt(twoWeeks)
             .exec(function(err, commits){
-                // if (err) {
```
seveibar: So is the error being handled? Shouldn't a different response be returned?

PR #46, opened 2015-03-28
Project Blog - updates #41

kmcnellis:
updates #41 , addresses #30


seveibar:
@agundy can you verify this is working?

client/app/projects/blog/blog.controller.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
+angular.module('observatory3App')
+.controller('ProjectsBlogCtrl', function ($scope, $http, $stateParams) {
+    $scope.load = function() {
```
agundy: Is this something that needs to be run like this? Not sure the function is needed.

server/routes.js
```diff
@@ -9,6 +9,7 @@ var errors = require('./components/errors');
 module.exports = function(app) {
 
   // Insert routes below
+  app.use('/api/posts', require('./api/post'));
```
agundy: Slightly nitpicking but not sure this is the best endpoint name. Would much rather have a blog endpoint and have posts under that.

agundy:
Works except you can start writing a post without being authenticated. This should not work until you login.

server/api/post/post.controller.js
```diff
@@ -0,0 +1,117 @@
...
+
+  req.body.author = req.user._id;
+  if (!req.body.projectId){ return res.status(400).send("Project not set"); }
+  Project.findOne({'_id': req.body.projectId }, function (err, project) {
+    if(err) { return handleError(res, err); }
+    if(!project) { return res.status(404).send("Project not found"); }
+
+    // Only someone who is part of the project can write a blog post
+    User.findById(req.user._id, function(err, user) {
+      console.log("user",user);
```
agundy: Shouldn't be in the commit

client/app/projects/blog/blog.controller.js
kmcnellis: The function reloads the blog posts after sending a new one in

server/routes.js
kmcnellis: I agree - @jfucci might have some insight?

server/api/post/post.controller.js
kmcnellis: What do you mean?

client/app/projects/blog/blog.controller.js
agundy: Okay cool. That makes sense. Just wasn't sure and didn't look at it long enough to realize that.

server/api/post/post.controller.js
agundy: Console logginging on the serve isn't necessary.

kmcnellis: Slipped through -  I'll fix it

server/api/post/post.controller.js
```diff
@@ -0,0 +1,117 @@
...
+    return res.json(post);
+  });
+};
+
+// Creates a new post in the DB.
+exports.create = function(req, res) {
+
+  req.body.author = req.user._id;
+  if (!req.body.projectId){ return res.status(400).send("Project not set"); }
+  Project.findOne({'_id': req.body.projectId }, function (err, project) {
```
agundy: This works but for future reference findById is built into Mongoose and does the same query with a little less code and arguably more clarity. http://mongoosejs.com/docs/api.html#model_Model.findById

agundy:
Thanks a bunch @kmcnellis and @jfucci! Merged.

PR #56, opened 2015-04-06
Fixed a dependency issue that was causing the collapsable navbar to not work correctly

jfucci:
@agundy should be happy about this fix!


agundy:
The build server doesn't seem to think it can find the right version

PR #61, opened 2015-04-07
home page changes

Frannycesca:
Created changes to the home page and connected features in the server seed and the client


Frannycesca:
I looked closer at what I sent in the pull request. Yes, in one of the new commits I did add the new API end points, but in a later one, I deleted them, so they are not actually included in the pull request.

Issue #59, opened 2015-04-07
Quick add project from URL


agundy:
Streamline project input with simple URL paste.

PR #65, opened 2015-04-08
Users can now "quick add" a project using a github url. Also revised the add project modal to look better.

jfucci:
Addresses #59


agundy:
Works great. Thanks!

PR #69, opened 2015-04-14
Added attendance checks in case already added and switched to database i...

agundy:
...tem

### DIFF
--- a/server/api/user/user.controller.js
+++ b/server/api/user/user.controller.js
@@ -5,22 +5,22 @@ var passport = require('passport');
 var config = require('../../config/environment');
 var jwt = require('jsonwebtoken');
 var Commit = require('../commit/commit.model');
+var Thing = require('../thing/thing.model');
 var mongoose = require('mongoose');
 var request = require('request');
 
-
 var validationError = function(res, err) {
-  return res.json(422, err);
+    return res.json(422, err);
 };
 
 /**
  * Get list of users
  */
 exports.index = function(req, res) {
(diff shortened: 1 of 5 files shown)


server/api/user/user.controller.js
```diff
@@ -5,22 +5,22 @@ var passport = require('passport');
...
 
 /**
  * Get list of users
  */
 exports.index = function(req, res) {
-  User.find({},'-hashedPassword -salt -github.events -attendance -tech', function (err, users) {
-    if(err) return res.send(500, err);
-    res.json(200, users);
-  });
+   User.find({}, '-salt -hashedPassword', function (err, users) {
```
kmcnellis: Why did you return the attendance, github events, and tech to the user list?

server/api/user/user.controller.js
```diff
@@ -340,44 +336,43 @@ exports.authCallback = function(req, res, next) {
...
-            $push: {
-                attendance: new Date()
-            }
-        }, function(err){
-            if (err) return res.send(500, err);
-            res.send({"success":(err !== 0)});
-
-        });
-    }
+
```
kmcnellis: This should verify that the daycode is still valid (same day, or time range of dates)

agundy: We currently don't have any time stamp on the daycode. But it only adds the date that it was entered so if we haven't replaced the day code yet and they submit on a different day it will show the day submitted.

PR #73, opened 2015-04-15
Working file upload and static hosting

natedawg1013:

PR #74, opened 2015-04-15
Dependency fixes to pass the travis.ci build - Fixes #73

jfucci:


jfucci:
Fixes #73

server/api/project/project.controller.js
```diff
@@ -76,3 +79,36 @@ exports.destroy = function(req, res) {
...
+    var file = files.file[0];
+    var name = file.path.substring(file.path.lastIndexOf('/')).substring(1);
+    var cwd = process.cwd();
+    var path = cwd + '/server/static/uploads/' + req.params.username + '/' + req.params.project;
+    var destPath = path + '/' + name;
+    if(!fs.existsSync(path)){
+      mkdirp.sync(path);
+    }
+    fs.rename(file.path, destPath, function (err) {
+      if (err) console.error(err)
```
agundy: These should return errors to the clients. Console errors on the server are just going to be eaten in the logs and never seen

server/routes.js
```diff
@@ -15,13 +16,15 @@ module.exports = function(app) {
   app.use('/api/things', require('./api/thing'));
   app.use('/api/users', require('./api/user'));
 
+  app.use('/uploads', express.static( __dirname + "/static/uploads" ));
```
agundy: Could we move this outside of the project directory? the current deployment system relies on the ability to move the entire project structure but photos shouldn't be reset between versions.

bower.json
```diff
@@ -15,10 +15,15 @@
     "lodash": "~2.4.1",
     "angular-ui-router": "~0.2.10",
     "angular-bootstrap": "~0.11.0",
-    "codemirror": "~5.1.0"
+    "codemirror": "~5.1.0",
+    "angular-file-upload": "~1.1.5",
+    "ng-file-upload": "~3.3.1",
+    "ng-file-upload-shim": "~3.3.1"
```
agundy: Are we actually using all three of these? Couldn't tell from the code but just want verification

PR #93, opened 2015-08-11
Reverted to uploading to file directory and switched to stream copy

agundy:


seveibar:
Do we use cwd in other places? This would mean the script can only be executed at the root of the directory. Is there a way to reference the root?

agundy:
cwd is the best way to reference the root. also running with grunt will always launch it from the root directory

seveibar:
:shipit:

PR #94, opened 2015-08-12
Forgot Password full implementation

### LINKED ISSUE #75: Password Reset
We should set up a password reset link

seveibar:
Fixes #75 

This code review is probably a good example as to how to make any general feature, as I had to touch nearly every part of the code.

### DIFF
--- a/server/api/user/user.controller.js
+++ b/server/api/user/user.controller.js
@@ -357,7 +357,7 @@ exports.resetPassword = function(req, res){
         email: userEmail.toLowerCase()
     }, function (err, user){
         if (err) return res.json(401, err);
-        if (!user) return res.json(401, {message: 'User does not exist'});
+        if (!user) return res.send(200);
 
         crypto.randomBytes(12, function(ex, buf) {
             var token = buf.toString('hex');
--- a/server/api/user/user.controller.js
+++ b/server/api/user/user.controller.js
@@ -4,6 +4,8 @@ var User = require('./user.model');
 var passport = require('passport');
 var config = require('../../config/environment');
 var jwt = require('jsonwebtoken');
+var crypto = require('crypto');
+var email = require("../../components/email");
 var Commit = require('../commit/commit.model');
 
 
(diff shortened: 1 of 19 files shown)


server/config/environment/production.js
```diff
@@ -13,6 +13,13 @@ module.exports = {
             process.env.PORT ||
             8080,
 
+  // Server Address
+  addr:     process.env.SERVER_ADDR ||
+            "",
+
+  // Sendgrid email
+  sendgridApiKey: process.env.SENDGRID_API_KEY,
```
agundy: this doesn't have to be in the index and the other environments? Why not only have this code in environment/index.js?

server/api/user/user.controller.js
```diff
@@ -340,3 +347,43 @@ exports.removeTech = function(req,res){
...
+
+            user.save(function(err){
+                if (err) return validationError(res,err);
+                res.send(200);
+
+                // email token to user
+                email.send(userEmail,
+                    "Observatory3 password reset",
+                    "Please go to the following url to reset your password\n\n\
+                    " + config.addr + "/login?token=" + user.passwordResetToken + "\n\n\
```
agundy: Make this an html hyperlink? Not just a string url?

server/api/user/user.controller.js
```diff
@@ -340,3 +347,43 @@ exports.removeTech = function(req,res){
...
+/**
+ * Set reset token for user and email it the password token will expire after 24 hours.
+ */
+exports.resetPassword = function(req, res){
+    var userEmail = req.body.email;
+    User.findOne({
+        email: userEmail.toLowerCase()
+    }, function (err, user){
+        if (err) return res.json(401, err);
+        if (!user) return res.json(401, {message: 'User does not exist'});
```
agundy: Should not let this information escape to the client. Make every email have a check email option. We should also remove the email is not registered and make it a generic 'email/password incorrect'

agundy:
few minor things. Otherwise looks good and functions as expected. Thanks!

server/config/environment/production.js
seveibar: for development it's only on localhost:9000

seveibar: it's set there in the development.js environment file

server/api/user/user.controller.js
seveibar: not totally sure it would render properly- i've seen sites email just the link (some clients don't allow html?). Obviously something that should be looked at in the future

server/api/user/user.controller.js
seveibar: There is no password entered here- the only way to invoke the error is to not have the email exist (so it doesn't really matter what the error message is- they have the code as this is an OS project)

I think the solution to this is to have a captcha, which can be a separate issue.

server/config/environment/production.js
agundy: The sendgridapi key is in every file. Not needed

seveibar: in every file? Could you clarify?

seveibar:
creates #99 to address forgot password information reveal

server/config/environment/production.js
seveibar: so actually this variable should be defined for production but doesn't have to be defined for development, which is why development falls back to empty string but production doesn't fall back

seveibar: fixed

agundy:
Should also run jshint on this. you have some errors

PR #100, opened 2015-08-13
Prettied up the login and register pages a tad

agundy:


seveibar:
the login, register, forgot password buttons now look a bit weird

seveibar:
same thing with signup, the alignment should be right and not left

seveibar:
after fixing looks good to merge

seveibar:
on second thought, only thing is forgot password alignment
selection_006

seveibar:
looks good as is

Issue #104, opened 2015-08-17
"I work on this!" button


seveibar:
Button on projects page to allow user to add themselves to the project.

Issue #22, opened 2015-03-15
Profile Function Needs to return Correct Data


agundy:
Profile is currently serving some fake data for the front end. Need to update it with actual models later.

agundy:
Does the profile page have fake data on it still?

seveibar:
The only data that is "Fake" is attendance data, so this will be fixed after that merge (should be done in 20-40 min)

seveibar:
Moving this to the second milestone because attendance score isn't necessary for the first release

seveibar:
changed to #138 because it's more relevant

PR #107, opened 2015-08-17
User page enhancements and ability to join/leave project

### LINKED ISSUE #101: Clicking username in nav bar should redirect to user's profile page

seveibar:
Fixes #101 #104 

Probably the biggest step towards #22.

User can now join projects and a user's profile actually displays project information for projects they're on.

Notes:
* new api endpoints were created for projects.
* `/me` will now take you to your profile
* There's a new concept of being ON a project and being an author- I believe that users should be able to join a project without being an author, then author's can add people on the project to the author's group after they've made some amount of contribution (as determined by authors). This eliminates the problem where people have to wait for approval to join a project or someone wants to demonstrate interest in contribution without the commitment of fully joining the project. This could be confusing to project maintainers however because of the distinction between author's and people on the project. This is up for discussion.

### DIFF
--- a/client/app/projects/profile/profile.controller.js
+++ b/client/app/projects/profile/profile.controller.js
@@ -2,15 +2,24 @@
 'use strict';
 
 angular.module('observatory3App')
-.controller('ProjectsProfileCtrl', function ($scope, $http, $stateParams, $upload) {
+.controller('ProjectsProfileCtrl', function ($scope, $http, Auth, $stateParams, $upload) {
     $http.get('/api/projects/'+ $stateParams.username + '/' + $stateParams.project).success(function(project){
-      $scope.project = project;
+        $scope.project = project;
+        Auth.isLoggedInAsync(function(loggedIn){
+            if (loggedIn){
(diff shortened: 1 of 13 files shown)


agundy:
I think the concept of a follower of the project would be much more familiar and work better for our purposes instead of the distinction betwen author and on a project.

seveibar:
I think no matter what we need to do away with the concept of an "author"-
contributor is a much better term with less ambiguity.

Let's flip that term in the schema. Creating this whole following thing is
redundant with github and that's missing the point. Contributors should be
auto-generated and a user's projects field should have the projects theyve
joined so it shows what they're active in.

I think that's the best way to go about this.

agundy:
I really do not like the heavy usage of alerts. We need to make a notification system

agundy:
Alright. I'm just not sure the way you set up being on a project is intuitive

seveibar:
Yea the alerts are 100% a temporary measure- it should be a separate issue
because so much code already uses it.

PR #114, opened 2015-08-30
Admins can now change users roles.

### LINKED ISSUE #87: Add way to make person admin/mentor

agundy:
Fixes  #87.

### DIFF
--- a/client/app/user/profile/profile.controller.js
+++ b/client/app/user/profile/profile.controller.js
@@ -8,10 +8,12 @@ angular.module('observatory3App')
       Auth.getCurrentUser(function(user){
         loggedInUser = user;
         $scope.isuser = loggedInUser._id === $stateParams.id;
+        $scope.loggedInUserRole = loggedInUser.role;
       });
 
       $http.get('/api/users/' + $stateParams.id).success(function(user){
           $scope.user = user;
+          $scope.originalRole = user.role;
           $http.get('/api/commits/user/' + user.githubProfile).success(function(commits){
               $scope.user.commits = commits;
           });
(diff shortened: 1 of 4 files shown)


client/app/user/profile/profile.controller.js
```diff
@@ -8,10 +8,13 @@ angular.module('observatory3App')
       Auth.getCurrentUser(function(user){
         loggedInUser = user;
         $scope.isuser = loggedInUser._id === $stateParams.id;
+        $scope.loggedInUserRole = loggedInUser.role;
       });
 
       $http.get('/api/users/' + $stateParams.id).success(function(user){
           $scope.user = user;
+          $scope.originalRole = user.role;
+          console.log($scope.originalRole);
```
seveibar: log line

agundy: good catch

seveibar:
Can users be admins and mentors? I see a possible problem arising...

seveibar:
I think it's fine as long as we only have the 3 roles "user","mentor" and "admin", but if we have more we're going to have to do this in a more "tag-like" fashion

agundy:
I think we can assume if someone is an admin they are also some form of mentor

seveibar:
right, but we may want to introduce a new role e.g. faculty

seveibar:
like i said, it's fine for now, but possibly an issue later

agundy:
Agreed. But i just built this off our existing work. I don't think a refactor of role is a concern right now.

seveibar:
after the console.log change :shipit:

seveibar:
I agree

PR #115, opened 2015-08-30
initial attendance implementation

seveibar:
This is the initial attendance implementation, allows admins to generate attendance codes and for users to input them via a navbar button.

I'm actively working on this so some changes will be pushed over the next 20 minutes or so


server/api/user/user.controller.js
```diff
@@ -295,17 +296,30 @@ exports.authCallback = function(req, res, next) {
...
-    var result = User.update({
+    var code = req.body.dayCode;
+    if (!code) res.send(400, "No Code Submitted");
+    // Check code against current class year
+    ClassYear.findOne({
+      current: true
+    }, function(err, classYear){
+      if (err) return res.send(500, err);
+      if (!classYear) return res.send(500, "No Class Year");
+      if (classYear.dayCode === code){
```
agundy: different daycode per class year?

seveibar: The class year represents the current semester of rcos, all day codes are recorded along with some other information.

A lot of this is confusing because the utility methods aren't in

seveibar:
:shipit:

PR #126, opened 2015-09-01
Padding Change

rickrizzo:
Minor change, but fixes mobile performance so that there is no weird bar on the side of the site due to overflowing text.


rickrizzo:
Screenshot of error
screenshot_2015-09-01-12-42-17

agundy:
Looks good. Thanks!

PR #129, opened 2015-09-04
Removed old build instructions and switched to container builds

agundy:


agundy:
Since uploads were changed this should no longer need to be in the travis file or readme

seveibar:
Nice- now for the new stuff :)

PR #132, opened 2015-09-11
Create LICENSE.txt

kmcnellis:


agundy:
Should be combined with the other license PR

kmcnellis:
Does it matter? Done from the web interface.

agundy:
Fine :P

LICENSE.txt
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
```
seveibar: we should really be GPL

Issue #141, opened 2015-09-15
Blank user profile looks bad


agundy:
Need a way to make the user profile page look better when user hasn't filled out any info.

seveibar:
Similar to looking at an empty github repository, a message should be displayed like "This user hasn't added any tags" or something of the like

nickkoul:
Do you want anything other then default text before anything has been filled out?

seveibar:
Things like tags should display "This user has not placed any tags"

PR #151, opened 2015-09-24
Default User Profile

nickkoul:
@seveibar added default bio message and default tag message. Is this what you were looking for?

### DIFF
--- a/server/api/user/user.controller.js
+++ b/server/api/user/user.controller.js
@@ -257,9 +257,6 @@ exports.changeBio = function(req,res){
 
     User.findById(userId, function(err,user){
         user.bio = newBio;
-        if(newBio == ""){
-          user.bio = "No Bio written yet. Please tell us some more about yourself";
-        }
         res.json({bio:user.bio});
         user.save(function(err){
             if (err) return validationError(res,err);
--- a/server/api/user/user.controller.js
+++ b/server/api/user/user.controller.js
@@ -197,7 +197,7 @@ exports.destroy = function(req, res) {
 };
 
 /**
- * Change what role the user is 
+ * Change what role the user is
  * restriction: 'admin'
  */
 exports.role = function(req, res) {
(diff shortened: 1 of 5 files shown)


server/api/user/user.model.js
```diff
@@ -18,9 +18,9 @@ var UserSchema = new Schema({
   hashedPassword: String,
   provider: String,
   salt: String,
-  tech: [String],
+  tech: [String], 
   projects: [{type : Schema.Types.ObjectId, ref: 'Project'}], // project id
-  bio:String,
+  bio: {type:String, default: "No Bio written yet. Please tell us some more about yourself"},
```
seveibar: @agundy should we do this client side or server side? My intuition says it might be better on the client, but this does solve the problem.

server/api/user/user.controller.js
```diff
@@ -257,10 +257,15 @@ exports.changeBio = function(req,res){
 
     User.findById(userId, function(err,user){
         user.bio = newBio;
+        if(newBio == ""){
```
seveibar: hmm I feel like this could get messy (because finding users by id in other places might have a different bio) again, @agundy

server/api/user/user.model.js
agundy: I think it's much better to keep the check client side. That way we know
easily who hasn't update there info, all we have to do is check for an
empty string instead of querying/checking for a default string.

On Thu, Sep 24, 2015 at 3:19 PM, Severin Ibarluzea <notifications@github.com

seveibar:
Hey Nick! I think we're going to want to move the bio profile client-side, i.e. do a check in angular on the profile page.

nickkoul:
Sounds good I will change that

agundy:
Thanks! Shouldn't be hard. Just makes more sense for maintainability. Let
us know if you need help using angular.

seveibar:
Looks great! Tested and exactly what we were looking for. Thanks Nick!

kmcnellis:
Fixes #141

PR #154, opened 2015-09-26
Added post delete

zlalvani:
I did thing


seveibar:
Nice! Great bug fix on the server post controller!

Issue #139, opened 2015-09-15
URP Form enable/disable


seveibar:
Admins should be able to enable or disable the URP button at the top of the page.

seveibar:
fixed, nice job @leej42

PR #158, opened 2015-09-26
Show and Hide URP Form from admin page

leej42:
Create an API call for displayURP and getDisplayURP
Called the API in the navbar to show/hide the URP Form
Added a field to the classyear Model
Added user=admin to delete class year data

### DIFF
--- a/client/components/navbar/navbar.controller.js
+++ b/client/components/navbar/navbar.controller.js
@@ -1,7 +1,7 @@
 'use strict';
 
 angular.module('observatory3App')
-  .controller('NavbarCtrl', function ($scope, $location, Auth) {
+  .controller('NavbarCtrl', function ($scope, $location, $http, Auth) {
     $scope.menu = [{
       'title': 'Home',
       'link': '/'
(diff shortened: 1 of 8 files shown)


leej42:
fixes #139

client/components/navbar/navbar.controller.js
```diff
@@ -27,4 +27,17 @@ angular.module('observatory3App')
...
       return route === $location.path();
     };
+
+    // Toggles the display of URP form
+    var URPDisplay = function(){
+      $http.get("/api/classyear/displayURP")
+        .success(function(data){
+          $scope.displayURP = data.displayURP;
+        });
+        console.log("did it");
```
agundy: We should remove these console.log lines

agundy:
This does not fail correctly if there is no class year defined. You should add in a default if the lookup fails. Shouldn't make a difference in production but matters for development.

seveibar:
class year's will never not be defined- correction is a low priority issue

seveibar:
create #168 to address the issue, merging

PR #162, opened 2015-09-27
Updated upload and added deletions.

agundy:
Big changes. No promises... Made the upload workflow better with no need to refresh page. Also added functionality to delete the images.


agundy:
Change from Monsters to a little more professional images and add space between images on the project page. Then it's good to go! Thanks!

kmcnellis:
'x' button does not disappear after removing the last photo

kmcnellis:
note: requires a bower install

PR #177, opened 2015-10-02
Fixed URL ending without / and added build in deploy script

agundy:
Build change shouldn't break anything for now as I updated it on the server but for future changes to packages it's needed. Also no sure what happened but the projects url changed slightly without being updated throughout. Perhaps just a package change? Tested on live Observatory and works.


agundy:
@seveibar no review?

seveibar:
I'll review it after my nap

agundy:
thanks for the :shipit:

Issue #217, opened 2015-10-09
If no commits for a user than show a message saying no commits found


agundy:
This should just be a ng-show check against the user.commits.length and only show the message if the count is 0. The user profile code can be found `/client/app/user/profile` and you want to look at the profile.controller.js and the profile.html

forestballad:
Anli & Sumit are working on this

chocolatemelt:
Kevin & Tae are working on this (Table 6)

PR #222, opened 2015-10-09
Added ng-show check against user.commits.length

parkt2:
Issue #217


agundy:
Could you move this to inside the commit box? It's around line 84 of the same file. Thanks!

agundy:
This looks like it didn't fail because of you.

agundy:
One test passed and the other timed out. I will look into it but you are fine! Thanks.

PR #226, opened 2015-10-09
tested password in user.spec.js

helei96:
Is this good enough?


agundy:
That looks like the start of what we want. Good enough for the lab. I know what's broken, thanks!

PR #229, opened 2015-10-10
Add RSS Feeds to projects

### LINKED ISSUE #183: RSS Feeds For Blogs
I made a fancy blog for my RCOS project and already post regularly, I'd love to just drop the RSS link in the Blog page and not have to worry about it anymore

burkel24:
Fixes #183 
Adds Support for RSS feeds on projects. Also adds support for 'external' blog links to an arbitrary url (not embedded like rss).

Check out [this screenshot](http://imgur.com/6yb5uZk) for an idea of how it works.

This commit also moves the editing of project details into a modal, which feels like a much cleaner design to me. See [this link](http://imgur.com/xHLVczZ) for a preview

Question: Is there a precommit / linting script I should run before merging


agundy:
Looks good but when adding the external blogging you completely removed the internal blogging implementation. We had basic markdown blog post support previously.

agundy:
There is npm test which is run by travis and you can run grunt jshint for linting, though the current project isn't really following that too well.

burkel24:
I figured that if they were using an external blog they probably wouldn't care to use the built in. Do you think it should support both?

Hmm would you be interested in a more beefy precommit type script that would do linting and style enforcement. I noticed a lot of style fragmentation throughout the project

agundy:
Is there not a way to switch between them though? I couldn't quite figure that out. All I saw were options for adding an external blog.

blivingston24:
I could add a toggle but it might be confusing to have so many blogs, unsure. I could remove the external blog type entirely, add a way to let project owners toggle or just display links to both. Which sounds right to you?

agundy:
I think we can just have RSS feeds or Observatory integrated blog, any basic blogging platform has RSS integration so I wouldn't worry about the plain external link until people ask for it.

kmcnellis:
Any updates on this @agundy @burkel24?  It would be nice to get this merged (although there are now conflicts)

burkel24:
refactoring due to the load of conflicts, maybe tomorrow

agundy:
@kmcnellis what is the "Need clarification" label supposed to mean?

kmcnellis:
It means that there is some problem holding back the pull request from being merged. It's especially useful when looking through the list to see that most have been looked at and asked for something to change rather than having to re-review all each time. Ideally the author would remove the tag after fixing

vkr96:
@burkel24 how is this going?

vkr96:
should someone get assigned to do this? or should we rebuild this from the ground up?

PR #233, opened 2015-10-18
Hide add button

### LINKED ISSUE #216: Hide add project button from users who are not logged in
Also needs to be taken care of in the back end (as per issue #161) but the front end change will be a good start

vkr96:
Simple front-end change to hide_add_button. Someone else wrote the authentication line, so I am simply pushing this front-end change again. 

fixes #216


client/app/projects/projects.controller.js
```diff
@@ -32,6 +47,10 @@ angular.module('observatory3App')
         }
     };
 
+   $scope.checkUserProject = function() {
```
agundy: Not quite sure what is going on here, could you comment it?

client/app/projects/projects.controller.js
```diff
@@ -1,15 +1,30 @@
...
+            $scope.checkUserProject();
+        }
+    });
 
     $scope.getCurrentProjects = function() {
         $http.get('/api/projects').success(function(projects) {
             $scope.projects = projects;
         });
         $scope.past = false;
+        Auth.isLoggedInAsync(function(loggedIn){
```
agundy: Why do this again? Why not just set the loggedIn flag in the other function that you call on page load?

client/app/projects/projects.html
```diff
@@ -5,7 +5,7 @@
     <div id="header">
     <h1> {{ past ? "Past" : "Current" }} RCOS Projects</h1>
 
-    <button type="button" class="btn btn-primary" data-toggle="modal" data-target="#addProject">
+    <button type="button" class="btn btn-primary" data-toggle="modal" data-target="#addProject" ng-show = "userOnProject">
```
agundy: Can you remove the spaces between ng-show, =, and the string? HTML generally keeps things attached with no spaces between assignment of variables.

vkr96:
This should fix it

PR #236, opened 2015-10-18
Commit length

nickkoul:
fixed default commit length


seveibar:
Looks great thanks Nick!

PR #238, opened 2015-10-18
Added footer to bottom of every page 

### LINKED ISSUE #169: All pages need footers
The footer on the homepage should be on ever page via an include like the navbar.

vkr96:
Footers for all pages like the one on the main page 

fixes #169


client/components/footer/footer.controller.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+angular.module('observatory3App')
+  .controller('MainCtrl', function ($scope, $http) {
```
agundy: This shouldn't be called MainCtrl. And doesn't look like a controller is even needed at this point. You could leave it out for now. We can add it back when we need more logic/functionality

agundy:
I also see you adding the footer too a component and adding it to other pages but you left it on the main page. You should remove the footer from the main page and put in the include so that we only have to update the footer in one location.

vkr96:
I'm so sorry for the disjointed commits.....

seveibar:
Looks good to me! Any last thoughts @agundy?

kmcnellis:
The footer is still missing on the user profile, and the signup/login pages, it looks like

vkr96:
That should fix the problem

kmcnellis:
Could you also add a link to the github in the footer?

vkr96:
If you click on issues it leads to the github

PR #239, opened 2015-10-18
Added 'default' marker

### LINKED ISSUE #136: RCOS "Starter/Default Projects"
Admins should be able to mark projects in RCOS as "Starter" or "Default", these projects are intended for users who just join RCOS to join.

There should be a button on the projects page "Mark Default"/"Unmark Default" that toggles a project's status.

zlalvani:
projects can now be marked as default, fixes #136


client/app/projects/profile/profile.controller.js
```diff
@@ -4,6 +4,9 @@
 angular.module('observatory3App')
 .controller('ProjectsProfileCtrl', function ($scope, $http, Auth, $stateParams, Upload, Project, notify) {
     $scope.userOnProject = false;
+
+
```
agundy: a little too much added whitespace?

agundy:
I think I missed something but what will this marked as default functionality be used for? Projects that people can be auto added to?

zlalvani:
I think the idea is to let users join a project that has been marked as default when they register.

zlalvani:
fixes #136

seveibar:
@agundy this was part of our idea to prevent people from not having a project

PR #244, opened 2015-10-23
Automatically adds the creator of a project as a member of that project

### LINKED ISSUE #218: Adding a project should also add the user creating it as a contributor
There should be a call added to /server/api/project/project.controller.js in the exports.create function to add the project to the user's project list (similar to exports.addProject in /server/api/user/user.controller.js)

bennyty:
This (theoretically, I'm not very familiar with your structure) should automatically add the creator of a project as a member of that project. Please comment if I've done something wrong. #CSCI-2961 Lab 6

fixes #218


server/api/project/project.controller.js
```diff
@@ -54,6 +54,22 @@ exports.authors = function(req, res) {
 exports.create = function(req, res) {
   Project.create(req.body, function(err, project) {
     if(err) { return handleError(res, err); }
+
+    //Add creator to any new project
+    var userId = req.params.id;
```
agundy: I think you want req.user._id. Params doesn't have the user id in it at this point. Did you test it?

PR #249, opened 2015-10-26
Display attendance information correctly on profile page and keep track of user's bonus days

### LINKED ISSUE #138: Display attendance information correctly on profile page

### LINKED ISSUE #228: Make it so that the user can only submit the day code once
Right now a user can submit the same code twice and an error is thrown that crashes the server.
The backtrace shows that line 360 in server/api/user/user.controller.js is where the error is thrown.

kmcnellis:
Look at the first 2 commits to see the pre-jshint changes.

Fixes #138 

also fixes #228

### DIFF
--- a/client/app/user/profile/profile.controller.js
+++ b/client/app/user/profile/profile.controller.js
@@ -2,18 +2,18 @@
 'use strict';
 
 angular.module('observatory3App')
-  .controller('ProfileCtrl', function ($scope, $stateParams, $http, Auth, notify) {
+    .controller('ProfileCtrl', function ($scope, $stateParams, $http, Auth, notify) {
 
-      var loggedInUser;
-      Auth.getCurrentUser(function(user){
-        loggedInUser = user;
-        $scope.isuser = loggedInUser._id === $stateParams.id;
-        $scope.loggedInUserRole = loggedInUser.role;
-      });
+        var loggedInUser;
+        Auth.getCurrentUser(function(user){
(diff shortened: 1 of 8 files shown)


seveibar:
eek! bonusAttendance should be a virtual! This will break attendance!

seveibar:
we can detect bonusAttendance days from the ClassYear

seveibar:
i.e. ClassYears keep track of bonusAttendance days, so you can reference a users attendance from bonusDays

seveibar:
other than that- this is good refactor

seveibar:
Small Group contains the fix for #228 already (it was lodged in one of the early commit messages)

agundy:
I think we should try to do linting changes in separate pr's much cleaner and easier to merge worry free

kmcnellis:
I was planning on it, but my editor did some by accident. Also, seve - fixing 228 it was a side effect of making the bonus days work. I'll refactor once your code is merged

kmcnellis:
And for the bonus days - its cleaner so that activities that happen to be on the same day as a meeting can have a bonus code.  (Potentially we should also do that in general, so you can have more than one code per day)

seveibar:
:+1: let's merge after small groups

seveibar:
this WILL still break live site attendance- the result will be those who attended a bonus day will get a valid day of attendance

vkr96:
@kmcnellis how is this going?

PR #264, opened 2015-10-30
User can now delete himself

vkr96:
Added a button for the user to delete themselves.


kmcnellis:
There should be a check before deletion to ask for confirmation "Are you sure you want to delete yourself?"

server/api/user/index.js
```diff
@@ -32,5 +32,6 @@ router.delete('/:id/project', auth.isAuthenticated(), controller.removeProject);
 router.put("/:id/addTech", auth.canEdit(), controller.addTech);
 router.put("/:id/removeTech", auth.canEdit(), controller.removeTech);
 router.post('/resetPassword', controller.resetPassword);
+router.put('/:id/removeUser',auth.isAuthenticated(),controller.destroy);
```
seveibar: hehe, this will allow any logged in user to delete another user

Also, where is `controller.destroy` defined?

kmcnellis:
Also, are we sure we want this on the profile? It should probably be hidden in settings.

seveibar:
And we may want to keep a history of a user- so maybe just a make inactive button?

seveibar:
Maybe only admins can delete?

vkr96:
The issue was made because a user wanted to delete a duplicate account. Having admins only be able to delete the account would go against what the users want

agundy:
I agree with Seve that we should actually remove it from the db. No need to delete data just hide it. Flagging as inactive should be a fine resolution. If you want you could make a third option of hidden. 



—

vkr96:
Isn't this code deleting it from the db currently?

agundy:
That may be what it was doing initially but if we are rewriting part of it we might as well make it how we want it.

agundy:
I also think we want this in the settings page.

vkr96:
I'm sorry for all the commits....

agundy:
don't be sorry.. fix it with `git rebase -i`

server/config/environment/development.js
```diff
@@ -17,5 +17,5 @@ module.exports = {
   // For testing verification
   // attendanceVerificationRatio: 1,
 
-  seedDB: false
+  seedDB: true
```
agundy: remove this change please

server/api/user/user.controller.js
```diff
@@ -577,3 +577,30 @@ exports.removeProject = function(req,res){
...
 };
+/*
+Function that is called by removeUser api call 
+
+*/
+exports.deleteUser = function(req,res,next){
+  var userId = req.params.id;
+  var pass = String(req.body.oldPassword);
+  var query = {students:{ $in: [userId]}};
+  console.log("hello");
```
agundy: extra log line. Please remove

server/api/user/user.controller.js
```diff
@@ -577,3 +577,30 @@ exports.removeProject = function(req,res){
...
+  User.findById(userId, function (err, user,db) {
+    if(user.authenticate(pass)) {
+       SmallGroup.findOneAndUpdate(query, {$pull: {address: userId}}, function(err, data){
```
agundy: are you sure you are removing from the address field? aren't you removing from the list of students?

server/api/user/user.controller.js
```diff
@@ -11,6 +11,7 @@ var ClassYear = require('../classyear/classyear.model');
 var SmallGroup = require('../smallgroup/smallgroup.model');
 
 
+
```
agundy: remove extra added line please?

server/api/user/user.controller.js
```diff
@@ -34,7 +35,7 @@ exports.index = function(req, res) {
 exports.search = function(req, res){
     if (!req.query.query) return res.send(400, "No query supplied");
     var query = new RegExp(["^", req.query.query, "$"].join(""), "i")
-    User.findOne({name: query}, function(err, user){
+    User.findOne({name: query},{active:{$nin:false}},function(err, user){
```
agundy: what is this doing here? You shouldn't be updating search code for deleting a user should you?

server/api/user/user.controller.js
```diff
@@ -290,6 +291,22 @@ exports.changePassword = function(req, res, next) {
...
 /**
+ * Deactivates a user
+ */
+exports.deactivate = function(req,res) { 
+        var userId = req.user.id;
+        User.findById(userId, function(err, user){
+        if (err) return res.send(500, err);
+        user.active = false;
+        user.save(function(err){
+        if (err) return res.send(500, err);
```
agundy: tab spacing is off here. Please fix

server/api/user/index.js
```diff
@@ -33,5 +33,7 @@ router.delete('/:id/project', auth.isAuthenticated(), controller.removeProject);
 router.put("/:id/addTech", auth.canEdit(), controller.addTech);
 router.put("/:id/removeTech", auth.canEdit(), controller.removeTech);
 router.post('/resetPassword', controller.resetPassword);
+router.put('/:id/removeUser',controller.deleteUser); 
+router.put('/:id/deactUser',auth.canEdit(),controller.deactivate);
```
agundy: we already have a deactivate route. See '/:id/deactivate' i think this is from a merge conflict you tried resolving.

client/app/user/profile/profile.html
```diff
@@ -33,6 +33,9 @@
                 <br/>
                 {{user.email}}
         </div>
+            <div ng-show='isuser'>
```
agundy: maybe allow mentors/admins to do this too

client/app/user/profile/profile.controller.js
```diff
@@ -85,6 +85,20 @@ angular.module('observatory3App')
           });
       };
 
+      $scope.pastUser = function(user){
+          var conf = confirm("Are you sure you want to make your account inactive?"); 
+          if (conf){
+            $http.put('/api/users/' + $stateParams.id + '/deactUser', {
```
agundy: Should make use of the User factory instead of calling $http. You may need to add the deactivate functionality to the factory though

agundy:
Went pretty hard on the comments but most of my comments are easy fixes. Let me know if you need any clarification @vkr96  Thanks for the awesome work!

client/app/user/profile/profile.html
vkr96: They already can from the admin panel

vkr96:
Whats wrong with the $http call?

agundy:
factory's are reusable and put all api endpoints in one location.

client/app/user/profile/profile.html
agundy: works for me

agundy:
Looks great! Thanks @vkr96

PR #269, opened 2015-11-04
Fixed some bugs with image uploads

### LINKED ISSUE #263: Uploading an image doesn't update the project page
should be an easy fix

jfucci:
This should be quick to go through and merge.

Fixes #263. Also fixes a bug I found where a project could have multiple images and deleting one would always delete the first image.


zlalvani:
Is the stuff deleted in profile.controller.js no longer necessary?

jfucci:
It was simply duplicate code... look at lines 43 - 65 of the same file. I don't know how or when it was copied but it seemed like a good a time as any to delete

agundy:
Probably my excess code. I rewrote it. Thanks for cleaning up!



—

PR #276, opened 2015-11-07
Adding user spec test

Istyatur:
Trying to figure out createUser api call

Still working on that api call

/auth/createUser?

Changing url to /api/user

Typo

Typo

Typo

Changing expected error type from 00 to 422 (unprocessable entity)

Debugging

removed log files and cleaned up branch

### DIFF
--- a/server/api/user/user.spec.js
+++ b/server/api/user/user.spec.js
@@ -4,7 +4,6 @@ var should = require('should');
 var app = require('../../app');
 var request = require('supertest');
 
-
 describe('GET /api/users/past', function() {
 
   it('should respond with JSON array', function(done) {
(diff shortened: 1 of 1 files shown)


server/api/user/user.spec.js
```diff
@@ -19,3 +18,23 @@ describe('GET /api/users/past', function() {
       });
   });
 });
+
+
+describe('POST /api/users/', function() {
+    it('should respond with JSON array', function(done) {
```
nickkoul: I think the exports.create function should respond with a token

seveibar:
@Istyatur can you fix this?

Istyatur:
I lack the knowledge to do so.

seveibar:
Look at the return type for creating a new user

seveibar:
@agundy how did this test even pass?

seveibar:
oh we took out the test condition

server/api/user/user.spec.js
```diff
@@ -19,3 +18,23 @@ describe('GET /api/users/past', function() {
...
+	    "user":{}
+	};
+
+	request(app)
+	    .post('/api/users/')
+	    .send(postData)
+	    .expect(422)
+	    .end(function(err, res) {
+		if(err) return done(err);
+		//res.body.should.be.instanceof(Array);
```
seveibar: let's change this to a todo to get this out of the PR queue

`//TODO check if token present`

server/api/user/user.spec.js
seveibar: `it('should respond with a token')`

PR #282, opened 2015-11-07
Updated google analytics to something we are tracking instead of the default

agundy:


seveibar:
cool give me the credentials for this sometime

PR #288, opened 2015-11-13
added search bar for projects

nickkoul:
Created a search bar on projects page that lets you search through current projects

### DIFF
--- a/client/app/projects/projects.html
+++ b/client/app/projects/projects.html
@@ -14,7 +14,12 @@
     <button type="button" class="btn btn-primary" ng-click="getPastProjects()" ng-show="!past">
         Show Past Projects
     </button>
+    <search-bar>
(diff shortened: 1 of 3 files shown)


client/app/projects/projects.html
```diff
@@ -118,7 +123,7 @@ <h4 class="modal-title" id="addProjectLabel">Add a Project</h4>
         </div>
     </div>
 
-    <div ng-repeat="project in projects">
```
seveibar: this is awesome- if somebody wants to know how to use a filter we should reference this PR

client/app/projects/projects.scss
```diff
@@ -78,6 +80,14 @@ h2 {
     float: left;
 }
 
+#header search-bar{
```
seveibar: I'm creating an issue to restyle this a bit but overall very well done

PR #267, opened 2015-11-03
Developers can now change whether a project is a past or current project

vkr96:
Fixes Issue #256

PR #294, opened 2015-11-21
Console error fixes

agundy:
Cleaning house


kmcnellis:
Fixed in #267 and #302

PR #295, opened 2015-11-22
Updated project page

### LINKED ISSUE #290: Restyle projects search bar
The search bar introduced in #288 doesn't look to great, let's restyle it

agundy:
Search box is now matching bootstrap theme and cleaned up design and made more mobile responsive.
Should fix #290

### DIFF
--- a/client/app/projects/projects.html
+++ b/client/app/projects/projects.html
@@ -2,25 +2,54 @@
 
 <div class="container">
 
-    <div id="header">
-    <h1> {{ past ? "Past" : "Current" }} RCOS Projects</h1>
-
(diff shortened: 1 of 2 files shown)


jfucci:
This looks good to me. The only think I notice is that the search magnifying icon acts like a button... might not be the best design choice since its a live search

kmcnellis:
I agree that the magnifying icon shouldn't be a button - also, the lines around the box are misaligned.  Otherwise its good

kmcnellis:
I'll pull for now, but added a new issue for a better search bar styling #307

PR #308, opened 2015-11-28
Display attendance information correctly on profile page and keep track of user's small group / bonus days 

### LINKED ISSUE #138: Display attendance information correctly on profile page

kmcnellis:
fixes #138, #262, #336


server/api/user/user.model.js
```diff
@@ -90,90 +108,409 @@ function isoDateToTime(isoDate){
...
         }
-      }
-      this.attendance.push(today);
-    }else if (status === "unverified"){
-      // If user already has attendance remove their attendance
-      for (var i = this.attendance.length-1;i >= 0;i--){
-        if (isoDateToTime(this.attendance[i]) == today.getTime()){
-          this.attendance.splice(i,1);
+        //New style attendence
+        var classYear = global.currentClassYear;
```
seveibar: the addition below these lines is incredibly verbose- i suspect this could have been done in a better way...

server/api/user/user.model.js
```diff
@@ -90,90 +108,409 @@ function isoDateToTime(isoDate){
...
+                            if (status === 'presentBonus' && thisDate.bonus === true && thisDate.smallgroup === false){
+                                found = true;
+                            }
+                            if (found){
+                                return;
+                            }
+                          }
+
+                        }
+                        if (status === 'present'){
```
seveibar: I think this level of complexity will make maintenance very difficult

seveibar:
There's some really good stuff in here, e.g. i like the new schema, but I can't help but feel that we should have a design discussion surrounding this because of the additional layer of complexity.

We also won't be able to merge until the end of the semester in order to prevent the current system from breaking, we'll also have to make a migration plan and get some documentation to explain this.

Additionally there is virtually no commenting on how this system works, there is a set of essentially enum strings that are undocumented.

I can't help but feel we might want to do this in smaller PRs, @agundy thoughts?

kmcnellis:
This will not actually break the current system - if you notice, it adds new stuff to the new system, but still reads the old one for all information.

kmcnellis:
To be fair, the old system was also undocumented.  I can add some comments, though.  It all has to work together, so multiple pull requests would result in many things not working / having to be written multiple times. If it had been put in place before small groups, that could have been a possibility.

client/app/user/profile/profile.scss
```diff
@@ -1,124 +1,135 @@
...
+        font-size: .75em;
+        padding-top: .25em;
+    }
 
-#attendance-box-container{
+    #attendance-box-container{
+        display: flex;
+
+        div{
+            // width:16px;
```
seveibar: ghost code (not a big deal)

kmcnellis:
And what emun strings?

server/api/user/user.model.js
kmcnellis: I was trying to leave flexibility in there.  As I worked in different parts, there are additional functions / properties that could be changed, but this was the original scheme.  It is intended to allow codes of all times to be generated on any one calendar day.

seveibar:
presentSmall, presentBonus etc.

Doing those as strings is not the cleanest way- it wouldn't be a big deal but it seemed to increase the backend logic.

And that's good that you maintained backwards compatibility (I hadn't noticed that) but in this case I would actually suggest not maintaining the old system to reduce the overall system complexity.

kmcnellis:
This was also meant to start the conversation. I would welcome more people commenting and making commits to the branch in order to further improve it.  We just needed someplace to start.

seveibar:
That's perfect thanks- although I might start the conversation with less than 1200 LOC changes ;)

kmcnellis:
For maintaining the old system - yes, I would agree once the semester is over

kmcnellis:
There was a lot of things that needed to be fixed in order to display on the fronted :P That was the issue, after all...

seveibar:
okay I see your game- but since we're so close to the end right now, I would just remove compatibility now. Attendance will hardly matter after this Friday

seveibar:
Fair enough, for clarification my big concerns ATM are...
* Remove backwards compatibility so we don't have to clean it up
* Add commenting and/or more modularity, there's a lot of huge scary pieces of code that is not broken up
* All these strings don't seem clean to me at all... if you see me in person please pitch it to me :)
* Possibly an document on the wiki explaining how this works a bit (I can help you with this)

agundy:
@seveibar and @kmcnellis are we done resolving this? And could we make it able to be merged?

agundy:
when we go to merge this could we reduce the commit count for this? I realize it shouldn't be all one commit but I think we should be able to reduce from the 24 commits it's currently sitting at.

kmcnellis:
That was the plan!

scripts/migrateAttendance.js
```diff
@@ -0,0 +1,178 @@
+// This file is to help migrate the old database which had users with an 'attendance' field
+// to a new attendance collection
+
+// var users = db.users.find({});
+//
+// while(users.hasNext()) {
```
agundy: no need to comment this out. why not remove it? We have it in history

scripts/migrateAttendance.js
```diff
@@ -0,0 +1,178 @@
...
+            db.users.update(
+                { "_id": user._id },
+                {
+                    $unset: { attendance: 1, unverifiedAttendance: 1 }
+                }
+            );
+            print("Migrated attendance for "+user.name+", attendamce unset");
+        }
+        else{
+            print("Failed to migrated all attendance for "+user.name+", attendamce not unset");
```
agundy: nit attendance spelled wrong

scripts/migrateAttendance.js
```diff
@@ -0,0 +1,178 @@
...
+
+                                verified: true,
+                                code: daycode.code,
+                            }
+                        );
+                        found = true;
+                        break;
+                    }
+                }
+                if (!found && user.smallgroup && smallgroup && smallgroup.dayCodes){
```
agundy: Comments would be good in here. Lots of conditionals to keep track of

scripts/migrateAttendance.js
```diff
@@ -0,0 +1,178 @@
...
+  date.setHours(0,0,0,0);
+  return date;
+}
+
+var smallgroups = db.smallgroups.find({});
+
+while(smallgroups.hasNext()) {
+    smallgroups.forEach(function(group){
+        if (group.semester){
+            var classyear = db.classyears.findOne({semester:group.semester})
```
agundy: do in callback instead of just inline?

scripts/migrateAttendance.js
```diff
@@ -0,0 +1,178 @@
...
+                var found  = false;
+                for (var b = 0; b < classyear.dayCodes.length ; b++){
+                    if (standardDate(classyear.dayCodes[b].date).getTime() == standardDate(date).getTime()){
+                        var daycode = classyear.dayCodes[b];
+                        db.attendances.insert(
+                            {
+                                classYear: classyear._id,
+                                user: user._id,
+
+                                date: standardDate(date),
```
agundy: time is still an important attribute for seeing if people are submitting later in the evening. Not sure we should strip it if we don't have too

scripts/migrateAttendance.js
```diff
@@ -0,0 +1,178 @@
...
+                        );
+                        found = true;
+                        break;
+                    }
+                }
+                if (!found && user.smallgroup && smallgroup && smallgroup.dayCodes){
+                    for (var c = 0; c < smallgroup.dayCodes.length ; c++){
+                        if (standardDate(smallgroup.dayCodes[c].date).getTime() == standardDate(date).getTime()){
+                            var daycode = smallgroup.dayCodes[c];
+                                db.attendances.insert(
```
agundy: indenting?

server/api/attendance/attendance.controller.js
```diff
@@ -0,0 +1,623 @@
...
+};
+exports.setAttendenceFullBonus = function(req, res) {
+    return setAttendenceInfo(req, res, false, true);
+};
+
+
+/**
+* Verify users large group attendance, available only to mentors
+*/
+// router.post('/present/:user', controller.verifyFullAttendance);
```
agundy: hasRole('mentor')

server/api/attendance/index.js
```diff
@@ -0,0 +1,37 @@
...
+
+var router = express.Router();
+
+router.get('/', auth.isAuthenticated(), controller.index);
+router.get('/:id', auth.isAuthenticated(), controller.show);
+router.get('/present/me', auth.isAuthenticated(), controller.getAttendanceMe);
+router.get('/present/me/:date', auth.isAuthenticated(), controller.presentMe);
+router.post('/attend', auth.isAuthenticated(), controller.attend);
+
+router.post('/', auth.hasRole('mentor') || auth.hasRole('admin'), controller.create);
```
agundy: as stated verbally. no need to explicitly allow `auth.hasRole('admin')` admin inherit all lower auth levels and can do anything a mentor can do. These should all be changed to `router.post('/', auth.hasRole('mentor'), controller.create);
 `

server/api/classyear/classyear.controller.js
```diff
@@ -27,6 +27,20 @@ exports.getClassYear = function(req, res) {
...
 };
 
+// Get a specific class year's attendance bonus days
+exports.countBonusDays = function(req, res) {
+  ClassYear.findOne({
+    "current": true
+  }, function (err, classYear){
+    if(err) { return handleError(res, err); }
+    if (!classYear) return res.send(404);
+    res.json(classYear.dayCodes.reduce(function(previousValue, currentValue, index, array) {
```
agundy: might be nice to break this into multiple lines and comment on what it is doing with the reduce. People new to js will have a hard time understanding this

server/api/classyear/classyear.controller.js
```diff
@@ -60,7 +74,10 @@ exports.create = function(req, res) {
           }
         }
       });
-      res.send(204);
+      ClassYear.getCurrent(function(err, currentClassYear){
+          global.currentClassYear = currentClassYear;
```
agundy: is this used by any other code?

server/api/post/post.controller.js
```diff
@@ -45,7 +45,7 @@ exports.create = function(req, res) {
     if(!project) { return res.status(404).send("Project not found"); }
 
     // Only someone who is part of the project can write a blog post
-    User.findById(req.user._id, function(err, user) {
+    User.findById(req.user._id,  '-salt -hashedPassword', function(err, user) {
```
agundy: the -salt and -hashedPassword should be hidden by the model and you shouldn't need to explicitly removed

server/api/smallgroup/index.js
```diff
@@ -14,7 +14,7 @@ router.put('/:id', auth.hasRole('mentor') || auth.hasRole('admin'), controller.m
 router.get('/:id/members', auth.isAuthenticated(), controller.getSmallGroupMembers);
 router.put('/:id/member', auth.hasRole('mentor') || auth.hasRole('admin'), controller.addMember);
 router.delete('/:id/member/:memberId', auth.hasRole('mentor') || auth.hasRole('admin'), controller.deleteMember);
-router.get('/:id', auth.isAuthenticated(), controller.getSmallGroup);
+router.get('/:id', auth.hasRole('mentor') || auth.hasRole('admin'), controller.getSmallGroup);
```
agundy: change to auth.hasRole('mentor')

server/api/user/index.js
```diff
@@ -18,20 +18,18 @@ router.get('/:id/avatar', controller.avatar);
 router.delete('/:id', auth.hasRole('admin'), controller.destroy);
 router.post('/:id/role', auth.hasRole('admin'), controller.role);
 router.get('/me', auth.isAuthenticated(), controller.me);
-router.get('/unverified', controller.getUnverifiedAttendanceUsers);
 router.put('/:id/password', auth.isAuthenticated(), controller.changePassword);
 router.put('/:id/deactivate', auth.isAuthenticated(), controller.deactivate);
 router.put('/:id/activate', auth.isAuthenticated(), controller.activate);
 router.get('/:id', controller.show);
+router.get('/:id/private', auth.canEdit() || auth.hasRole('admin'), controller.privateProfile);
```
agundy: canEdit should allow admins to edit

server/api/user/user.controller.js
```diff
@@ -205,23 +219,68 @@ exports.create = function (req, res, next) {
...
-    if (!user) return res.send(404);
-    res.json(user.profile);
+  User.findById(userId)
+  .select( '-salt -hashedPassword -passwordResetToken -passwordResetExpiration')
+  .populate('smallgroup')
+  .exec(function (err, user) {
+    if (err) {return next(err);}
+    if (!user) {return res.send(404);}
+    var profile = user.profile;
+    ClassYear.findOne({
```
agundy: should be get current

scripts/migrateAttendance.js
kmcnellis: Datetime keeps the time.  Date is for db querying ease

scripts/migrateAttendance.js
kmcnellis: This is for mongo console, so its not actually async

scripts/migrateAttendance.js
kmcnellis: We still keep this in datetime. Date is the date of the attendance, and for ease of db queries

server/api/classyear/classyear.controller.js
kmcnellis: Not right now, but I'd like it to be used everywhere (more efficient than asking the db constantly for the same data)

server/api/post/post.controller.js
kmcnellis: This is not yet the case, but it should be an issue

server/api/attendance/attendance.controller.js
```diff
@@ -0,0 +1,675 @@
...
+            if(err) { return handleError(res, err); }
+            return res.send(204);
+        });
+    });
+};
+// *******************************************************
+
+// *******************************************************
+// Creates a new attendance submission in the DB.
+// DON'T USE : does not generate required data for submission
```
agundy: why have this if this doesn't work correctly?

server/api/attendance/attendance.controller.js
```diff
@@ -0,0 +1,675 @@
...
+        if(err) { return handleError(res, err); }
+        return res.json(201, attendance);
+    });
+};
+// *******************************************************
+
+// *******************************************************
+// Updates an existing attendance submission in the DB.
+// DON'T USE : does not generate required data for submission
+// Restricted to admins
```
agundy: mentor?

server/api/attendance/attendance.controller.js
```diff
@@ -0,0 +1,675 @@
...
+// Get list of attendance submissions
+// Restricted to mentors
+// router.get('/', auth.isAuthenticated(), controller.index);
+exports.index = function(req, res) {
+    Attendance.find(function (err, attendance) {
+        if(err) { return handleError(res, err); }
+        return res.json(200, attendance);
+    });
+};
+// *******************************************************
```
agundy: is this a new standard you want to start enforcing?

kmcnellis: Sure

server/api/attendance/attendance.controller.js
kmcnellis: It is useful for people who know what they are doing, as an alternative to directly manipulating mongo.  It works, but doesn't check what you are doing (like the other version does)

server/api/attendance/attendance.controller.js
kmcnellis: Fixed. Forgot to change the string.

kmcnellis:
@agundy Final review?

client/app/admin/admin.controller.js
```diff
@@ -13,7 +13,38 @@ angular.module('observatory3App')
...
+                        }
+                        else if(submission.bonus){
+                            u.bonusAttendance +=1;
+                        }
+                        else{
+                            u.fullAttendance +=1;
+                        }
+                    }
+                }
+            }, function(){
```
agundy: how should we handle this error case?

client/app/admin/admin.controller.js
```diff
@@ -13,7 +13,38 @@ angular.module('observatory3App')
             // Use the User $resource to fetch all users
 
             $scope.users = [];
-            $scope.users = User.allstats();
+            User.allstats({},function(users){
+                $scope.users = users;
+
+                for (var a = 0; a < $scope.users.length; a++){
```
agundy: this processing could be simplified into a function like countUserAttendance() would be a lot cleaner imo

client/app/admin/admin.html
```diff
@@ -27,9 +71,12 @@
                     <option value="admin">Admin</option>
                 </select>
               </td>
-              <td><a href="http://github.com/{{user.github.login}}">{{user.github.login}}</a></td>
-              <td>{{ user.attendance }}</td>
-              <td>{{ user.github.events.length }}</td>
+              <td><a href="http://github.com/{{user.githubProfile}}">{{user.githubProfile}}</a></td>
```
agundy: should change href to ng-href

client/app/attend/attend.controller.js
```diff
@@ -4,20 +4,43 @@ angular.module('observatory3App')
...
+      }
+    };
 
     $scope.submitDayCode = function(){
       var user = Auth.getCurrentUser();
-      $http.put('/api/users/' + user._id + '/attend', {
+      $http.post('/api/attendance/attend', {
         dayCode: $scope.userDayCode
       }).success(function(info){
+        console.log(info);
```
agundy: extra console.log()

client/app/attend/attend.html
```diff
@@ -13,15 +13,54 @@
...
+        <hr />
+      </div>
+      <div class="col-sm-12">
+        <h3>Previous Attendance:</h3>
+      </div>
+      <div class="col-sm-12">
+        <table class="table table-hover">
+          <thead>
+            <tr>
+              <td ng-click="setSortOrder('datetime')">Date
```
agundy: at some point we might want to break this into a directive so that the directive can worry about putting the arrow on the header value but for now this is fine. Just a lot of duplicate code

client/app/mentor/mentor.controller.js
```diff
@@ -2,22 +2,23 @@
...
-            notify({ message: "Error Occurred: " + err, classes: ["alert-danger"] });
+                notify({ message: 'Error Occurred: ' + err, classes: ['alert-danger'] });
         	});
-    }
-    $scope.verifyAttendance = function(user){
-    	$http.put("/api/users/" + user._id + "/verifyAttendance")
+    };
+
+    $scope.verifyAttendance = function(submission){
+    	$http.put('/api/attendance/' + submission._id + '/verify')
```
agundy: single space vs two space tab?

client/app/user/profile/profile.controller.js
```diff
@@ -9,23 +9,117 @@ angular.module('observatory3App')
...
-              $scope.user.commits = commits;
-          });
-          // get all users projects information
-          $scope.projects = [];
-          user.projects.forEach(function(projectId){
-            $http.get("/api/projects/" + projectId).success(function(project){
-              $scope.projects.push(project);
+        //If you can access private data
+        if ($scope.isuser || $scope.loggedInUserRole === 'admin'){
+            $http.get('/api/users/' + $stateParams.id+'/private').success(function(user){
```
agundy: why aren't we using the user factory here?

client/app/user/profile/profile.controller.js
```diff
@@ -9,23 +9,117 @@ angular.module('observatory3App')
...
+                if (!$scope.user.bio){
+                    $scope.user.bio = "An awesome RCOS developer!";
+                }
+                $scope.originalRole = user.role;
+                //Total attendance is current attendance in large and small groups
+                var u = $scope.user;
+                u.full={normal:0, bonus:0};
+                u.small={normal:0, bonus:0};
+                u.unverified = 0;
+                for (var b = 0; b < u.attendance.length; b++){
```
agundy: this is repeat code from the admin page. Any way we could modularize this? Perhaps using a service that processes a users data fields

client/app/user/profile/profile.controller.js
```diff
@@ -9,23 +9,117 @@ angular.module('observatory3App')
...
+        else{
+            //Get non-private user information
+            $http.get('/api/users/' + $stateParams.id).success(function(user){
+                $scope.user = user;
+                if (!$scope.user.bio){
+                    $scope.user.bio = "An awesome RCOS developer!";
+                }
+                $scope.originalRole = user.role;
+
+                $http.get('/api/commits/user/' + user.githubProfile).success(function(commits){
```
agundy: since the commits aren't integrated yet I feel like you shouldn't even worry about calling a commit endpoint yet for any of this code

client/app/user/profile/profile.controller.js
agundy: also i think the private vs public data should at least call the same factory call if not the same endpoint. There is some code duplication between the private/public profile information that makes this unnecessarily long

client/app/user/profile/profile.controller.js
```diff
@@ -9,23 +9,117 @@ angular.module('observatory3App')
...
+                    }
+                }
+
+                $http.get('/api/commits/user/' + user.githubProfile).success(function(commits){
+                    $scope.user.commits = commits;
+                    });
+                // get all users projects information
+                $scope.projects = [];
+                user.projects.forEach(function(projectId){
+                  $http.get("/api/projects/" + projectId).success(function(project){
```
agundy: Use project service? https://github.com/rcos/Observatory3/blob/master/client/components/project/project.service.js

client/app/user/profile/profile.controller.js
```diff
@@ -9,23 +9,117 @@ angular.module('observatory3App')
...
+                }
+                $scope.originalRole = user.role;
+
+                $http.get('/api/commits/user/' + user.githubProfile).success(function(commits){
+                    $scope.user.commits = commits;
+                    });
+                // get all users projects information
+                $scope.projects = [];
+                user.projects.forEach(function(projectId){
+                  $http.get("/api/projects/" + projectId).success(function(project){
```
agundy: duplicates code for getting users project data above that could be condensed into one function use factory etc

client/app/user/profile/profile.controller.js
```diff
@@ -9,23 +9,117 @@ angular.module('observatory3App')
...
+                       $scope.user.commits = commits;
+                   });
+
+                  });
+                });
+            });
+        }
       });
 
+
```
agundy: unnecessary added spacing?

client/app/user/profile/profile.controller.js
```diff
@@ -9,23 +9,117 @@ angular.module('observatory3App')
...
+                $http.get('/api/commits/user/' + user.githubProfile).success(function(commits){
+                    $scope.user.commits = commits;
+                    });
+                // get all users projects information
+                $scope.projects = [];
+                user.projects.forEach(function(projectId){
+                  $http.get("/api/projects/" + projectId).success(function(project){
+                    $scope.projects.push(project);
+
+                   $http.get('/api/commits/user/' + user.githubProfile).success(function(commits){
```
agundy: don't worry about commits until we have the backend integrated with github. These endpoints could change

scripts/migrateAttendance.js
agundy: okay sounds good. Just checking

scripts/migrateAttendance.js
```diff
@@ -0,0 +1,179 @@
...
+                        );
+                        found = true;
+                        break;
+                    }
+                }
+                //If not found, find the correct attendance code in the smallgroup
+                //Check to make sure there is a smallgroup for the user with daycodes
+                if (!found && user.smallgroup && smallgroup && smallgroup.dayCodes){
+                    for (var c = 0; c < smallgroup.dayCodes.length ; c++){
+                        if (standardDate(smallgroup.dayCodes[c].date).getTime() == standardDate(date).getTime()){
```
agundy: this same long line is used in multiple places. why not make it a function that returns a boolean. would be easier to understand and update later.

server/api/attendance/attendance.controller.js
```diff
@@ -0,0 +1,674 @@
...
+function isoDateToTime(isoDate){
+    var date = new Date(isoDate);
+    date.setHours(0,0,0,0);
+    return date.getTime();
+}
+
+// *******************************************************
+// Get list of attendance submissions
+// Restricted to mentors
+// router.get('/', auth.isAuthenticated(), controller.index);
```
agundy: I don't really like having this line in the function declaration because it is subject to change and requires all developers remembering to change the function comment code if they change route names, or permissions. It also is not hard at all to have the route definitions open to at the same time which has this info in it.

server/api/attendance/attendance.controller.js
agundy: see above but this forgetting is why i don't think these lines should be in the controller definitions

server/api/attendance/attendance.controller.js
```diff
@@ -0,0 +1,674 @@
...
+};
+// *******************************************************
+
+// *******************************************************
+// Verify users small group attendance given submission Id, available only to mentors
+// TODO: verify that mentor is part of the smallgroup
+// router.post('/present/:user/small', auth.hasRole('mentor'), controller.verifySmallAttendance);
+exports.verifySmallAttendance = function(req,res){
+    var userId = req.params.id;
+    return ClassYear.getCurrent(function(err, classYear){
```
agundy: what's with these returns.

server/api/attendance/attendance.controller.js
```diff
@@ -0,0 +1,674 @@
...
+exports.getAttendance = function(req, res) {
+    var userId = req.params.user;
+    return ClassYear.getCurrent(function(err, classYear){
+        var classYearId = classYear._id;
+        getAttendance(userId, classYearId, function(userAttendance){
+            return res.json(userAttendance);
+        });
+    });
+};
+exports.getAttendanceMe = function(req, res) {
```
agundy: why is this needed? seems like a get attendance when not given a user should just return the logged in user. either that or require a user to pass in their own user id

server/api/attendance/attendance.controller.js
```diff
@@ -0,0 +1,674 @@
...
+    }
+    var userId = req.params.user;
+    return ClassYear.getCurrent(function(err, classYear){
+        var classYearId = classYear._id;
+        getPresent(userId, date, classYearId, function(userAttendance){
+            return res.json(userAttendance);
+        });
+    });
+};
+exports.presentMe = function(req, res) {
```
agundy: not sure i like these wrapper functions that just personalize the more generic function

server/api/attendance/attendance.controller.js
```diff
@@ -0,0 +1,674 @@
...
+// router.get('/present/:user/today', auth.hasRole('mentor'), controller.present);
+// router.get('/present/:user/:date', auth.hasRole('mentor'), controller.present);
+//
+// router.get('/present/me/today', auth.isAuthenticated(), controller.presentMe);
+// router.get('/present/me/:date', auth.isAuthenticated(), controller.presentMe);
+
+// Check if a userId is present today
+var getPresent = function(userId, date, classYearId, cb){
+    var callback = cb || function(){};
+    Attendance.find({user:userId, date: date, classYear:classYearId}, function (err, attendance) {
```
agundy: shouldn't this be a findOne?

server/api/attendance/attendance.controller.js
```diff
@@ -0,0 +1,674 @@
...
+exports.present = function(req, res) {
+    var date = req.params.date;
+    if (req.params.date === 'today'){
+        date = isoDateToTime(new Date());
+    }
+    else{
+        date = isoDateToTime(req.params.date);
+    }
+    var userId = req.params.user;
+    return ClassYear.getCurrent(function(err, classYear){
```
agundy: why did you start using these returns that return a promise? You have a later return which is the only item actually returned.

server/api/attendance/attendance.controller.js
```diff
@@ -0,0 +1,674 @@
...
+        if (err) {return res.send(500, err);}
+        return getPresent(user._id, isoDateToTime(new Date()), classYear._id, function(userAttendance){
+            var submittedFull = false;
+            var submittedSmall = false;
+            var submittedSmallBonus = false;
+            var submittedFullBonus = false;
+            // Check what types of attendance the user has submitted today
+            for (var a = 0; a < userAttendance.length ; a++){
+                if (userAttendance[a].bonusDay && !userAttendance[a].smallgroup){
```
agundy: this should be simpllifed with a function that takes the attendance and returns a type. Also shouldn't we only have one attendance item per day?

server/api/attendance/attendance.controller.js
```diff
@@ -0,0 +1,674 @@
...
+                    date: isoDateToTime(new Date()),
+                    datetime: new Date(),
+
+                    bonusDay: false,
+                    smallgroup: false,
+
+                    verified: !needsVerification,
+                    code: code,
+
+                    }, function(err,submission){
```
agundy: doesn't handle error on save condition

server/api/attendance/attendance.controller.js
```diff
@@ -0,0 +1,674 @@
...
+                    datetime: new Date(),
+
+                    bonusDay: true,
+                    smallgroup: false,
+
+                    verified: !needsVerification,
+                    code: code,
+
+                }, function(err,submission){
+                    return res.send(200, {'type':'Full group bonus attendance', 'unverified': needsVerification});
```
agundy: doesn't return error on save. Always returns a 200

server/api/attendance/attendance.controller.js
```diff
@@ -0,0 +1,674 @@
...
+                            date: isoDateToTime(new Date()),
+                            datetime: new Date(),
+
+                            bonusDay: true,
+                            smallgroup: true,
+
+                            verified: !needsVerification,
+                            code: code,
+
+                        }, function(err,submission){
```
agundy: again error condition never checked or handled

server/api/attendance/attendance.controller.js
```diff
@@ -0,0 +1,674 @@
...
+        });
+    });
+};
+// *******************************************************
+
+
+// *******************************************************
+// Set attendance as present (no verification)
+// router.post('/attend/small/:user', auth.hasRole('mentor'), controller.setAttendanceSmall);
+// router.post('/attend/full:user/', auth.hasRole('mentor'), controller.setAttendanceFull);
```
agundy: `full:user/` typo? these router comments appear to be unreliable.

server/api/attendance/attendance.controller.js
```diff
@@ -0,0 +1,674 @@
...
+                        }
+                        else{
+                            submittedSmall.verified = true;
+                            submittedSmall.save();
+                            return res.send(200, {'type':'Small group attendance', 'unverified': needsVerification});
+
+                        }
+                    }
+                    // if not, create the attendance object
+                    Attendance.create({
```
agundy: I've seen many of these Attendance.create functions. any way we could simplify this step so it isn't 10 loc every time?

server/api/attendance/attendance.controller.js
```diff
@@ -0,0 +1,674 @@
...
+
+                }
+            });
+        }
+    });
+
+};
+
+// Get data for submitting attendance, then pass it to setAttendance
+var setAttendanceInfo = function(req, res, small, bonus){
```
agundy: why can't setAttendance handle this logic? Why do we need 5 other functions just to call setAttendance?

server/api/attendance/attendance.model.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+var mongoose = require('mongoose'),
+    Schema = mongoose.Schema;
+
+var AttendanceSchema = new Schema({
+
+    classYear: {type : Schema.Types.ObjectId, ref: 'ClassYear', index: true},
```
agundy: Should we break these into multiline definitions for clarity like so? 
```
classYear: {
    type : Schema.Types.ObjectId, 
    ref: 'ClassYear', 
    index: true
}, 
```

server/api/attendance/index.js
```diff
@@ -0,0 +1,44 @@
...
+router.get('/present/me/:date', auth.isAuthenticated(), controller.presentMe);
+router.get('/present/:user/:date', auth.hasRole('mentor'), controller.present);
+router.get('/present/me', auth.isAuthenticated(), controller.getAttendanceMe);
+router.get('/present/:user', auth.hasRole('mentor'), controller.getAttendance);
+
+// Mark attendance as present, subject to verification
+router.post('/attend', auth.isAuthenticated(), controller.attend);
+
+// Set attendance as present (no verification)
+router.post('/attend/small/:user', auth.hasRole('mentor'), controller.setAttendanceSmall);
```
agundy: why do mentors have 4 different api endpoints but the user side of attend is only one? can't we slim this down?

server/api/attendance/index.js
```diff
@@ -0,0 +1,44 @@
...
+var auth = require('../../auth/auth.service');
+
+var router = express.Router();
+
+// Basic Access / Manipulation
+router.get('/', auth.hasRole('mentor'), controller.index);
+router.get('/:id', auth.isAuthenticated(), controller.show);
+router.delete('/:id', auth.hasRole('mentor'), controller.destroy);
+
+// DON'T USE : does not generate required data for submission
```
agundy: why is this here if it shouldn't be used?

server/api/classyear/classyear.model.js
```diff
@@ -5,13 +5,12 @@ var mongoose = require('mongoose'),
...
   semester: String,
-  bonusDays: [Date],
   current: false,
-  displayURP:{type: Boolean, default: true},
+  displayURP:{type: Boolean, default: false},
   mentors: [{type : Schema.Types.ObjectId, ref: 'User'}],
   students: [{type : Schema.Types.ObjectId, ref: 'User'}],
   projects: [{type : Schema.Types.ObjectId, ref: 'Project'}],
-  dayCodes: [{date:Date, code:String, bonusDay:Boolean}]
+  dayCodes: [{date:Date, code:String, bonusDay:{type:Boolean, default:false}}]
```
agundy: getting complex for just one line

server/api/classyear/classyear.model.js
```diff
@@ -39,5 +38,62 @@ ClassSchema.statics.getCurrent = function(cb){
...
+	.get(function(){
+		var today = new Date();
+		today.setHours(0,0,0,0);
+		for (var i = 0;i < this.dayCodes.length;i++){
+			if (this.dayCodes[i].date.getTime() === today.getTime()){
+				return this.dayCodes[i];
+			}
+		}
+		return null;
+	});
```
agundy: no spaces between functions?

server/api/classyear/classyear.model.js
```diff
@@ -39,5 +38,62 @@ ClassSchema.statics.getCurrent = function(cb){
...
+ClassSchema
+    .virtual("days")
```
agundy: tab issue

client/app/mentor/mentor.controller.js
kmcnellis: ?

client/app/admin/admin.controller.js
kmcnellis: What error case?

client/app/user/profile/profile.controller.js
kmcnellis: This was already in the code

client/app/user/profile/profile.controller.js
kmcnellis: This was already in there and not in the scope of this pr. You should file an issue if it should be changed.

client/app/user/profile/profile.controller.js
kmcnellis: This was already in there and not in the scope of this pr. You should file an issue if it should be changed.

client/app/user/profile/profile.controller.js
kmcnellis: This was already in there and not in the scope of this pr. You should file an issue if it should be changed.

client/app/user/profile/profile.controller.js
kmcnellis: This was already in there and not in the scope of this pr. You should file an issue if it should be changed.

server/api/attendance/attendance.controller.js
kmcnellis: It is much simpler to look at this way, than having to cross reference 2 files with the front end code. It also allows better searchability and removes the need to look at the index.html most of the time. You should be looking at the usage of the endpoint anyway if you are changing the permissions (and that does not ofter happen)

server/api/attendance/attendance.controller.js
kmcnellis: This was because it was the original creation of the endpoints and a typo. The mere fact that you could at a glance notice that the endpoint should not be for a mentor is proof that it is useful

server/api/attendance/attendance.controller.js
kmcnellis: It ensures correct exiting from the function and no write after ends or multiple res.sends.  It is much easier to verify that everything is correct and is one control thread (since there was no promises at the time of writing)

server/api/attendance/attendance.controller.js
kmcnellis: This is simply a wrapper to enable router.get('/present/me'). It is effectively the same as using the logged in user, but enables /present/me to have weaker permissions than /present/:user (since users can only get their own if not a mentor)

server/api/attendance/attendance.controller.js
kmcnellis: No. A user can have multiple attendances on the same day (full group, and a bonus activity, ect)

server/api/attendance/attendance.controller.js
kmcnellis: see above

server/api/attendance/attendance.controller.js
kmcnellis: see above

server/api/attendance/attendance.controller.js
kmcnellis: No, users can have a large group and a bonus activity in the same day. Or a small group and a bonus. (Say, a hackathon on a friday)

server/api/attendance/attendance.controller.js
kmcnellis: Its just a typo.... It is meant as documentation for the future. Thats why you are supposed to verify it.

server/api/attendance/attendance.controller.js
kmcnellis: It would be much more confusing to have a function take in 10 parameters...

server/api/attendance/attendance.controller.js
kmcnellis: For permissions / user friendliness. Trying to keep it cleaner than a giant function...

server/api/attendance/index.js
kmcnellis: See above. It shouldn't be used normally, but is still useful for overriding....

server/api/attendance/attendance.controller.js
kmcnellis: Plus, if you didn't notice. The comment is actually true to the router - it is a problem in the router itself

client/app/user/profile/profile.controller.js
kmcnellis: Not going to address in this pr

server/api/attendance/index.js
kmcnellis: Because we use the code to figure out the type of attendance

kmcnellis:
@agundy Addressed your comments

client/app/mentor/mentor.controller.js
agundy: looks like a spacing issue in your code.

client/app/user/profile/profile.controller.js
agundy: ugh what happened to leaving things prettier than you found them?

server/api/attendance/attendance.controller.js
agundy: I suppose so.

agundy:
Thanks Kiana, Looking better!

kmcnellis:
@agundy Whats holding this up now?

agundy:
I haven't done local testing yet. i will test and then upgrade

~Aaron Gunderson
Cell: 267-897-4308

PR #310, opened 2015-12-01
can now edit the small group name

### LINKED ISSUE #252: Small groups should be able to change name after being made

nickkoul:
Added field to edit and save the small group name. Can be done by anyone in the small group 
Fixes #252


kmcnellis:
This should probably be only enabled for mentors in the small group and admins

nickkoul:
Changed it so need proper permission to change the small group name

kmcnellis:
Thanks! I'll take a closer look and pull soon

kmcnellis:
I merged in the master branch, and it looks like there are some formatting/alignment inconsistencies with the show attendance button. 
screen shot 2015-12-03 at 4 28 21 pm
(its OK, just something to look into - maybe place the edit button next to the text rather than below it)

Also, the button to save should not be present when not editing.  When editing, it should then show the save button, as well as cancel.

server/api/smallgroup/smallgroup.controller.js
```diff
@@ -132,6 +132,20 @@ exports.addMember = function(req, res){
...
+exports.changeName = function(req,res){
+  var id = req.params.id;
+  var newName = String(req.body.smallGroupName);
+  SmallGroup.findById(id, function(err,smallgroup){
+    if (err) return handleError(res, err);
+    smallgroup.name = newName;
+    res.json({name:smallgroup.name});
+    smallgroup.save(function(err){
+      if (err) return validationError(res,err);
+      res.send(200);
```
kmcnellis: This causes an error on the server - line 141 (`res.json({name:smallgroup.name});`) should happen after saving (instead of the current line 144)  It should also be returned ( i.e., `return res.json. ..... `)

kmcnellis:
Is this ready? If it is, you should change the label to "status: ready to review"

agundy:
well it still has merge conflicts so i it can't be ready yet... @kmcnellis :stuck_out_tongue:

agundy:
@gitcole i merged the changes though I may have messed things up. Please check everything. Looks like you still have boxes around all the bootstrap elements. Small group adding works well though doesn't look great in my browser. Not sure if something you can style though. Also when a user is not a part of a small group the create a small group button is unstyled and doesn't match the other buttons

Issue #297, opened 2015-11-24
Padding on Project Website, Repo, Blog buttons


bennyty:
![](http://puu.sh/lxh4j/8ec5c80044.png)
These buttons need dental work.

PR #315, opened 2015-12-02
Moved tech tags to a more logical place

agundy:
some basic formatting fixes and moved tech tags to their own box instead of floating below everything else.

### DIFF
--- a/client/app/projects/profile/profile.html
+++ b/client/app/projects/profile/profile.html
@@ -5,7 +5,7 @@
         <div class="col-sm-3">
             <h1 ng-hide="edittingName">{{project.name}}</h1>
             <p>
-            <pname></pname>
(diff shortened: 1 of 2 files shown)


kmcnellis:
Related to #299 and #297

seveibar:
for client side/display fixes we can probably speed up the process by including a screenshot (as a general note)

kmcnellis:
+1

agundy:
well review it without a screenshot for now. All you have to do is checkout the branch :P I want to get this in so that we can update.

kmcnellis:
Still not a fan of the tech tags, but better than before. Fixes #297

agundy:
This doesn't fix #297...

kmcnellis:
It fixed it for me...

PR #321, opened 2015-12-06
Fix a bunch of issues with small groups

kmcnellis:
Includes daycare generation on demand, UI, permissions, deleting users. Also makes view of codes the same for class control and small groups.


kmcnellis:
@seveibar Could you take a look?

client/app/smallgroup/smallgroup.controller.js
```diff
@@ -2,78 +2,114 @@
...
                 }
-            })
+            });
         });
     };
 
-    $scope.showAttendance = function(){
-      if ($scope.showAttendanceCode){
+    $scope.generateAttendanceCode = function(){
+      if ($scope.showAttendanceCode){ //TODO call api to generate code
```
agundy: it looks like you added this functionality

agundy:
:shipit:

PR #332, opened 2016-01-29
Avoid mixed content; fixes #331

### LINKED ISSUE #331: Gravatar link protocol should match that used by the webpage
If a user is on https://rcos.io, Gravatar resources should also use HTTPS.

cyrozap:
This makes Gravatar URLs use whatever protocol the site uses.


kmcnellis:
thanks!

Issue #330, opened 2016-01-29
Day code submission should be case-insensitive


cyrozap:
A simple toUpperCase/toLowerCase should do the trick.

kochman:
I think this issue should be closed. PR #333 fixes this and was merged.

cyrozap:
I can confirm that this fixes the issue, thanks!

PR #333, opened 2016-02-01
Uppercase day codes so that they are case-insensitive

kochman:
Addresses #330. This bugged me the other day so I thought I'd fix it.


justetz:
Model citizen :clap: @kochman

kmcnellis:
Thanks for the fix!

PR #351, opened 2016-02-25
Added grunt seed command

agundy:
broke users and projects out into separate json files for easy additions. Also made it so adding a generator later wouldn't be too much additional work

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -593,6 +593,12 @@ module.exports = function (grunt) {
         }
       }
     },
+
+    execute: {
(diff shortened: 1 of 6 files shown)


jfucci:
looks like mocha can't find the 'promise' module that you are trying to use. Should probably fix that test before we merge

agundy:
added promise to the dependencies. should work now.

agundy:
@jfucci any other concerns now that the tests are passing?

kmcnellis:
Looks good!

Issue #347, opened 2016-02-23
Developer's icons should show a placeholder while they load


cyrozap:
It's a little jarring to watch each image appear randomly as they load, so it might be a good idea to have a loading placeholder using either CSS or a GIF (like something from [here](http://ajaxload.info/)).

agundy:
totally agree! Good tip! will fix. Thanks for the feedback @cyrozap

PR #352, opened 2016-02-25
Developers page updates

agundy:
- Now default to having a white background placeholder which fixes the
  moving asynch load of gravatars
- Adjusted the sizing of images on smaller screens

addresses #347

### DIFF
--- a/client/app/user/user.html
+++ b/client/app/user/user.html
@@ -13,7 +13,7 @@
             <div class="col-sm-6 text-left col-bottom col-height">
                 <div class="inside">
                     <span class="glyphicon glyphicon-chevron-left" aria-hidden="true" ng-click='decrement()'></span>
-                        {{ currentPage + 1 }} of {{ numberOfPages() }}
+                    {{ currentPage + 1 }} of {{ numberOfPages() }}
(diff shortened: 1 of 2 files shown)


jfucci:
looks like this failed due to a timeout... can someone else confirm this is just a travis issue? The code seems like it is fine

agundy:
yea this is a travis issue. The one worked and the other did not. Trying to find a way to rerun the checks

agundy:
Any opposition to me merging this?

kmcnellis:
@agundy It looks good, but you should drop the resolution of the gravitars as well

kmcnellis:
Also, I think 1 developer (full width, or with a small margin) looks better than 2 side by side on a small screen

agundy:
I changed that because 1 developer means that you have to scroll through 50 to get to the next page. And the avatars are huge.

kmcnellis:
2 side by side just looks odd, though.  Maybe try 3? or make sure there is some sort of margin

agundy:
I'll revisit it

agundy:
Updated!

PR #366, opened 2016-03-13
Fix grunt build

kmcnellis:
Looks like the problem was ` <!-- build:js({.tmp,client}) app/app.js -->` instead of  ` <!-- build:js({.tmp}) app/app.js -->`	.  The build script changed a bit, and the extra `,client` was causing stuff to be copied over twice for some reason.

### DIFF
--- a/.yo-rc.json
+++ b/.yo-rc.json
@@ -30,7 +30,7 @@
       "should": false,
       "expect": true
     },
-    "generatorVersion": "3.3.0",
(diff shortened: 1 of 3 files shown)


agundy:
can this go in before #308 ?

kmcnellis:
Yep- It probably should.

agundy:
Sweet tested and works for me. Merging. Thanks :)

Issue #255, opened 2015-10-27
The small group tab should not be visible to developers not in one


kmcnellis:
The api call `<li ng-show="isLoggedIn() && (isMentor() || hasSmallGroup())"><a ng-href='/smallgroup'>Small Group</a></li>
` does not seem to sufficiently prevent this

kmcnellis:
The error I'm getting (for a mentor) is specifically:
  var responseObject = smallgroup.toObject();
                                       ^
TypeError: Cannot read property 'toObject' of null
    at Promise.<anonymous> (/Users/kiana/dev/Observatory3/server/api/smallgroup/smallgroup.controller.js:59:40)

It shows the page but crashes the server.

kmcnellis:
For a student:
/Users/kiana/dev/Observatory3/server/api/smallgroup/smallgroup.controller.js:102
        smallgroup.students.forEach(function(studentId){
                  ^
TypeError: Cannot read property 'students' of null
    at Promise.<anonymous> (/Users/kiana/dev/Observatory3/server/api/smallgroup/smallgroup.controller.js:102:19)

kmcnellis:
Related to #354

PR #375, opened 2016-04-05
Allow users to see the small group

### LINKED ISSUE #354: All mentor-specific small group permissions can be executed by mentees
Mentees can remove group leaders and generate day codes if they go to `/smallgroup`

kmcnellis:
Fixes #354, #255


agundy:
looks like it needs to be rebased first

agundy:
Works! awesome.

Issue #361, opened 2016-03-08
View All Developers Bugged


gitcole:
When clicking view all on another other page other than page one the user has to navigate back to page 1 to view all users

PR #387, opened 2016-04-06
View all now resets page number to 1

bkovacik:
This fixes bug #361. I also fixed the disappearance of the "view less" button, which was bugged because someone added an extra condition to its display conditional.


agundy:
Hey works great! Thanks.

kmcnellis:
This is now bugged if you have too few users for a full page

PR #389, opened 2016-04-06
Prevent holding enter from submitting a multiple requests.

### LINKED ISSUE #365: Attend page should wait for a "key up" event before displaying another error message
If I press and hold "Enter" on the Attend page, I get a waterfall of error messages. To prevent users like me from doing this, the page should wait for a "key up" event before it evaluates the form submission again.

screenshot from 2016-03-11 16-25-10

agundy:
Fixes #365


kmcnellis:
It still allows the user to hold error and get a ton of already submitted requests back.

kmcnellis:
This should use angular's dirty marker on the form field

PR #390, opened 2016-04-06
Add time to blog posts and prevent user from setting time.

### LINKED ISSUE #367: Blog uses client system time
(I think) the blog system uses client time for the timestamps for the blog. Should this be the system time of the server, so that students cannot "spoof" the times in which they wrote the blogs?

agundy:
Fixes #367


kmcnellis:
Looks good!

PR #391, opened 2016-04-06
Fix delete user page

kmcnellis:
Change to model and fix styling.  Use DELETE instead of PUT

### DIFF
--- a/client/app/account/settings/confirmDelete/confirmDelete.directive.js
+++ b/client/app/account/settings/confirmDelete/confirmDelete.directive.js
@@ -10,7 +10,7 @@ angular.module('observatory3App')
       Auth.deleteUser($scope.password)
       .then(function(response){
         $scope.submitted = false;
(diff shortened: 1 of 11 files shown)


client/app/account/settings/confirmDelete/confirmDelete.directive.js
```diff
@@ -0,0 +1,32 @@
...
+          $scope.errors = {};
+
+          form.password.$setValidity('mongoose', false);
+          $scope.errors.password = 'Incorrect password';
+      });
+
+    }
+  };
+  $scope.close = function(){
+    // $uibModalInstance.close($scope.event);
```
agundy: ghost code

agundy:
Medium size page is much thinner than it needs to be: 
screenshot 2016-04-08 at 2 27 39 pm

kmcnellis:
It uses the same styling as the login page

Issue #102, opened 2015-08-17
Search users by tech


seveibar:
Search userbase by people who know X, where X is a tech present in the tech field of a user (specified on the profile page)

kmcnellis:
@vkr96 Any update on this?

nickkoul:
Is there functionality to search user by name already? If not should we do that as well?

aeksco:
Tracking this issue in [Observatory-Client #31](https://github.com/rcos/observatory-client/issues/31) - closing.

PR #392, opened 2016-04-06
User & Project page styling

### LINKED ISSUE #240: Search for users
There should be a search bar to find users

kmcnellis:
Make buttons & spacing similar
Add search to users

Fixes #240, works towards #102


agundy:
looks good

agundy:
how do we want to address #102? I think having it inline without seeing the tech would be confusing

kmcnellis:
We can add a button?

PR #393, opened 2016-04-06
Fixed server crash when updating github username.

agundy:
Also improved looks of sidebar


kmcnellis:
The ui changes don't look great on smaller screen widths 
screen shot 2016-04-08 at 9 19 16 am

kmcnellis:
And probably should hide the link while editing (make it in-place)
screen shot 2016-04-08 at 9 20 19 am

kmcnellis:
Enter does not submit changed github name.

kmcnellis:
Why can't email be updated the same way?

PR #401, opened 2016-04-12
Styling admin page and view bonus correctly

kmcnellis:


agundy:
mentioned in person but don't like the past and current user distinction. makes it hard if you accidentally click a user to be past. Also should try to find a font awesome icon instead of an x

gitcole:
There is a gylicon for that

kmcnellis:
Changed it to have check boxes. Let me know what you think!

kmcnellis:
screen shot 2016-04-15 at 10 42 37 am

agundy:
i don't think you need to have the activate/deactivate text below the checkbox

kmcnellis:
It is so that you can deactivate just 1 person immediately

kmcnellis:
screen shot 2016-04-15 at 1 25 08 pm

agundy:
Github links should open in a new tab. Also icons in the topbar should have title text to explain what they stand for, ie '+' > 'Bonus Days'

agundy:
I don't think we should have the names centered

PR #404, opened 2016-04-13
User profile

agundy:


kmcnellis:
Bio should be multi-line (and handle line breaks)

kmcnellis:
Should be able to click the words "Active Developer" to toggle the checkbox

agundy:
Addressed both comments

PR #406, opened 2016-04-13
Redesign and coloring

agundy:
Initial Design Overhaul for the homepage and other styling: 
See image for screenshot of homescreen
screenshot 2016-04-13 at 3 45 54 am 

Here is a picture of another page to show top bar coloring on different tabs:
screenshot 2016-04-13 at 3 47 26 am


kmcnellis:
Looks good! We should look at changing the primary color for buttons, though

PR #407, opened 2016-04-14
format settings.controller.js strings

GrahamAtkinson035:
All strings in settings.controller.js have the same format


agundy:
Thanks!

Issue #170, opened 2015-09-27
Email not working on live site


seveibar:
Just needs configuration tokens

agundy:
this works. Only problem is the email doesn't send the link correctly

agundy:
i got an email but it was to / instead of rcos.io/

kmcnellis:
The response is shown to the user in json 
![screen shot 2016-04-15 at 11 21 24 am](https://cloud.githubusercontent.com/assets/5566205/14566139/42786d32-02fc-11e6-9105-93d00cf80a1a.png)

PR #415, opened 2016-04-19
Updated forgot password to work with urls and the Sendgrid templates.…

agundy:
… Also fixed token login

### DIFF
--- a/client/app/account/forgot-password/forgot-password.controller.js
+++ b/client/app/account/forgot-password/forgot-password.controller.js
@@ -1,22 +1,21 @@
 'use strict';
 
 angular.module('observatory3App')
-  .controller('ForgotPasswordCtrl', function ($scope, Auth) {
-    $scope.user = {};
-    $scope.errors = {};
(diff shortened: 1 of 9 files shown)


agundy:
fixes #170

kmcnellis:
Not sending the correct error to the user when there is a error on the server

agundy:
Fixed error handling

client/app/account/forgot-password/forgot-password.controller.js
```diff
@@ -1,22 +1,22 @@
...
-    };
-
-  });
+    if(form.$valid) {
+      Auth.resetPassword($scope.user.email, function(err, res){
+        if (!err){
+          $scope.success = true;
+        }
+        else {
+          notify({ message: "Error: Problem Sending Password Reset Email", classes: ["alert-danger"] });
```
kmcnellis: Shouldn't be start with caps for every word

agundy:
updated error message

PR #420, opened 2016-04-19
Past and active users and projects can now be directly linked to

agundy:


kmcnellis:
The developer tab in the nabber is no longer highlighted when looking at past users

agundy:
Fixed

Issue #426, opened 2016-04-22
No Save Button on "Edit Profile" Page 


JohnBehnke:
There appears to not be a save button on the edit profile page. I'm using Safari. 

<img width="1093" alt="screen shot 2016-04-22 at 5 08 45 pm" src="https://cloud.githubusercontent.com/assets/5631869/14754979/e91260a2-08ac-11e6-9ae3-5dd0d12ab957.png">

agundy:
Something broke 😢 will look into this. Thanks for the issue

kmcnellis:
fixed in #432

PR #427, opened 2016-04-24
Fixed HTML comments to be right

agundy:
This broke grunt build.


agundy:
should fix #426

agundy:
still doesn't fix the save button bug.

PR #430, opened 2016-04-26
Fixed the fullscreen bug

gitcole:


kmcnellis:
Added some extra changes to make sure the code stays centered

PR #444, opened 2016-09-12
Email link

vkr96:
Changed one line to be https://rcos.io/


bkovacik:
Could you squash your commits?

server/api/user/user.controller.js
```diff
@@ -522,7 +522,7 @@ exports.resetPassword = function(req, res){
 
               var sub = {
                 ":name": [user.name],
-                "[%address%]": [config.addr + "/login?token=" + user.passwordResetToken],
+                "[%address%]": ["https://rcos.io/login?token=" + user.passwordResetToken],
```
agundy: Looks like this is actuallly just a config issue on the server. I'll dig into it but config.addr should be set already

PR #446, opened 2016-09-13
text-aligned right on developer page past button

jessamarie:


agundy:
Looks great! Thanks for the first pr! Ship it when green

PR #461, opened 2016-09-20
Added a filter by alphabetical order button to the projects page

jessamarie:


client/app/projects/projects.controller.js
```diff
@@ -13,6 +14,15 @@ angular.module('observatory3App')
         }
     });
 
+    $scope.getSortOrder = function(){
+
```
agundy: extra white line could we remove this?

client/app/projects/projects.controller.js
```diff
@@ -13,6 +14,15 @@ angular.module('observatory3App')
         }
     });
 
+    $scope.getSortOrder = function(){
```
agundy: maybe `toggleSortOrder()`? since it is actively changing the variable and not just returning a value?

agundy:
looks good. ship when you want!

PR #474, opened 2016-09-27
Added markdown to the achievements page

agundy:
that was easy

### DIFF
--- a/client/app/achievements/achievements.less
+++ b/client/app/achievements/achievements.less
@@ -1,3 +1,7 @@
 ng-invald {
   color: red
 }
+
+.event-description > img {
+  width: 100%;
+}
--- a/client/app/achievements/achievements.controller.js
+++ b/client/app/achievements/achievements.controller.js
@@ -32,5 +32,5 @@ angular.module('observatory3App')
         });
     };
 
-    $scope.isAdmin = Auth.isAdmin();
+    $scope.isAdmin = Auth.isAdmin;
   });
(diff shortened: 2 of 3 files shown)


agundy:
screenshot 2016-09-26 at 11 47 17 pm

client/app/achievements/achievements.less
```diff
@@ -1,3 +1,7 @@
 ng-invald {
   color: red
 }
+
+.event-description > img {
+  width: 100%;
```
Bad-Science: Should only top level images be fixed width?

Bad-Science: If so then LGTM

agundy: If the image is unstyled I don't want them blowing out of the box but if additonal styling is applied it should accept that. What case do you think this would break?

Bad-Science: An image within a link would break this and that case wouldn't break the layout. I think an image within a list would also break it, although that may cause it to blow out of the box  anyway so you can probably ignore that case.

agundy: Tested with a markdown image and it works as expected. Going to go with this for now
screenshot 2016-09-27 at 10 43 58 am

agundy: `<div id="content" class="event-description" btf-markdown="ach.description"><p><a href="http://google.com.au/">Foo</a></p></div>`

agundy:
thanks @Bad-Science

client/app/achievements/achievements.less
kmcnellis: Would the markdown styler ever apply extra styling to pictures? If not, then I think it is fair to add a max-width: 100% to all pictures. On the other hand, the width: 100% would certainly blow away any styling and should maybe be switched to max-width as well

PR #494, opened 2016-10-04
Added an alphabetical sort button to developer's page.

jessamarie:


agundy:
Did you test this with pagination?

client/app/user/user.html
```diff
@@ -47,7 +51,7 @@
       </div>
     </div>
     <div class="row image-row">
-        <div ng-repeat="user in users | filter:searchString | startFrom:currentPage*pageSize | limitTo:pageSize">
+        <div ng-repeat="user in users | filter:searchString | startFrom:currentPage*pageSize | limitTo:pageSize |  orderBy: sortOrder">
```
agundy: Watch out for the ordering here! you are limiting the length of the list and then sorting, sorting should happen first.

client/app/user/user.controller.js
```diff
@@ -23,6 +24,14 @@ angular.module('observatory3App')
       $scope.users = $scope.currentUsers;
     }
 
+    $scope.toggleSortOrder = function(){
+      if($scope.sortOrder === '-name'){
```
agundy: Nitpick but i like spaces between if and parentheses

jessamarie:
I didn't think to test this with pagination originally, but after moving the sort filter it's working on my end.

agundy:
All good! this is why we do code reviews :smile:

kmcnellis:
How would this work with the back button? Maybe we should add the sorting parameter to the url as well.

jessamarie:
@kmcnellis I'm on it!!

PR #495, opened 2016-10-05
Fix attendance so it does not use user.smallgroup

### LINKED ISSUE #257: Users can be in multiple small groups, but can only see one

### LINKED ISSUE #465: Small Group Attendance Calculator
Looks like there is a bug in small group attendance. Got this screenshot from Theo Rice. Looks like it is counting last semester attendance on the profile.

image

### LINKED ISSUE #483: Users that are in two small groups cannot attend the first one.
Possibly related to #257 

@seveibar I went to early/normal session (Sage 3713) instead of #latesession today btw.

kmcnellis:
Fixes #483
fixes #257
Should fix #465


client/app/smallgroup/smallgroup.controller.js
```diff
@@ -10,24 +10,18 @@ angular.module('observatory3App')
...
     callback = callback || function () {};
-
-    if (!$scope.user.smallgroup){
-      $scope.loaded = true;
-      $scope.smallgroup = null;
-      return;
-    }
-
-    $http.get('/api/smallgroup/' + $scope.user.smallgroup).success(function (smallgroup) {
+    return User.smallgroup({id:$scope.user._id})
```
agundy: did you get this working? You were asking about this before

client/app/smallgroup/smallgroup.controller.js
```diff
@@ -10,24 +10,18 @@ angular.module('observatory3App')
...
+    .$promise.then(function(smallgroup){
       $scope.smallgroup = smallgroup;
       $scope.loaded = true;
-      if (!smallgroup) {
+      if (!$scope.smallgroup._id) {
         $scope.smallgroup = false;
```
agundy: any reason this is a boolean? why not null?

client/app/smallgroup/smallgroup.controller.js
```diff
@@ -10,24 +10,18 @@ angular.module('observatory3App')
...
         return false;
       }
       if ('dayCode' in smallgroup && smallgroup.dayCode) {
         $scope.dayCode = smallgroup.dayCode;
       }
-      $http.get('/api/smallgroup/' + $scope.user.smallgroup + '/members').success(function (members) {
+      return $http.get('/api/smallgroup/' + $scope.smallgroup._id + '/members').success(function (members) {
```
agundy: do we have a factory for this yet?

client/app/smallgroup/smallgroup.controller.js
```diff
@@ -90,7 +83,7 @@ angular.module('observatory3App')
 
   $scope.saveSmallGroupName = function () {
     $scope.edittingSmallGroupName = false;
-    $http.put("/api/smallgroup/" + $scope.user.smallgroup + "/name", {
+    $http.put("/api/smallgroup/" + $scope.smallgroup + "/name", {
```
agundy: `$scope.smallgroup._id` since it's not the user attribute that is an id?

client/app/smallgroup/smallgroup.controller.js
```diff
@@ -115,13 +107,12 @@ angular.module('observatory3App')
   };
 
   var submitDayCode = function(code){
-      var user = Auth.getCurrentUser();
       $http.post('/api/attendance/attend', {
         dayCode: code
-      }).success(function(info){
+      }).success(function(){
```
agundy: ignoring the response from the API? things get sent back like "Please verify" do they not?

server/api/attendance/attendance.controller.js
```diff
@@ -335,17 +335,13 @@ exports.attend = function(req,res){
...
-        if (!user.smallgroup || user.smallgroup === undefined || user.smallgroup === null){
-          return res.send(400, 'No small group found/incorrect daycode!');
-        }
-
-        return SmallGroup.findById(user.smallgroup, function(err, smallgroup){
+        return SmallGroup.findOne({"students":user._id, "classYear":classYear._id}, function(err, smallgroup){
           if (err) {return handleError(err)}
           // if the user has no smallgroup, they cannont submit smallgroup  attendance
-          if (!smallgroup){return res.send(400, 'Incorrect day code');}
+          if (!smallgroup){return res.send(400, 'No small group found or incorrect daycode!');}
```
agundy: i think we are autoadding users to the smallgroup if they aren't part of one and they submit the right daycode. So Incorrect Daycode should be the message shouldn't it?

server/api/attendance/attendance.controller.js
```diff
@@ -335,17 +335,13 @@ exports.attend = function(req,res){
...
+
           // Small group, and not a bonus day
-          else if (smallgroup.dayCode === code){
+          if (smallgroup.dayCode === code){
```
agundy: reasoning for doing this? This changes expected behavior

server/api/project/project.controller.js
```diff
@@ -38,7 +38,6 @@ exports.stats = function(req, res) {
       ],
       function(err, results){
         if (err) {
-          console.log(err);
```
agundy: we should log errors shouldn't we? Maybe not to the console but somewhere

server/api/user/user.controller.js
```diff
@@ -60,7 +60,6 @@ exports.publicStats = function(req, res) {
       ],
       function(err, results){
         if (err) {
-          console.log(err);
```
agundy: api should log failures somewhere

server/api/user/user.controller.js
```diff
@@ -245,9 +244,7 @@ exports.show = function (req, res, next) {
     if (err) {return next(err);}
     if (!user) {return res.send(404);}
     var profile = user.profile;
-    ClassYear.getCurrent(function (err, classYear) {
```
agundy: weirddd why did we ever do this?

server/config/seed.js
```diff
@@ -23,15 +23,18 @@ var Project = require('../api/project/project.model');
 var Post = require('../api/post/post.model');
 var Smallgroup = require('../api/smallgroup/smallgroup.model');
 var ClassYear = require('../api/classyear/classyear.model')
+var Attendance = require('../api/attendance/attendance.model')
 
 
 // Seed the database with the original sample data
-var seed = function() {
+var seed = function(cb) {
```
agundy: I think seve changed this for a reason. @seveibar is this true?

server/config/seed/projects.json
```diff
@@ -15,7 +15,7 @@
     "githubProjectName": "yacs"
 },
 {
-    "_id": "5738cebdffd7ec3b75f3349b",
+    "_id": "000000000000000000000010",
```
agundy: doesn't seem necessary?

server/config/environment/development.js
```diff
@@ -17,5 +17,5 @@ module.exports = {
   // For testing verification
   // attendanceVerificationRatio: 1,
 
-  seedDB: true
+  seedDB: false
```
kmcnellis: Yup. I thought we had false as the default since the grunt seed command was created. It must have gotten switched at some point.

server/api/user/user.controller.js
kmcnellis: Not really sure..

client/app/smallgroup/smallgroup.controller.js
kmcnellis: Yes, this works

client/app/smallgroup/smallgroup.controller.js
kmcnellis: No particular reason, but this is the way it was before, and ensures that the front-end treats it correctly

client/app/smallgroup/smallgroup.controller.js
agundy: :+1:

client/app/smallgroup/smallgroup.controller.js
kmcnellis: This is no different from the original code. Small groups do not currently require verification

server/api/attendance/attendance.controller.js
kmcnellis: We are not yet doing that. That is a future issue.

server/api/attendance/attendance.controller.js
kmcnellis: It does not change the behavior. The previous if returns if true.

server/api/project/project.controller.js
kmcnellis: That would be a different issue

server/config/seed/projects.json
kmcnellis: It is - makes it much easier to write and verify tests

client/app/smallgroup/smallgroup.controller.js
kmcnellis: fixed

server/config/seed.js
kmcnellis: I just tried this without cb and the test failed

kmcnellis:
@agundy Not really sure why the test is failing. It works locally for me. Could you check it out?

server/config/seed.js
seveibar: this function should return a promise and should not have a callback- this is the new javascript paradigm

kmcnellis: I already changed it back. Not sure why github doesn't update the snippets anymore.

agundy:
Not sure why it's sending a 400. Failed once for me the second time I ran it. But hasn't failed again. Maybe a flakey test?

kmcnellis:
It should work now @agundy

agundy:
:shipit:

PR #498, opened 2016-10-05
Fix buttons and tech-tags

kmcnellis:
screen shot 2016-10-05 at 12 20 53 am
Suggestion for buttons
Also made tech tags consistant


kmcnellis:
@agundy Can you take a look at merging this into your PR?

PR #499, opened 2016-10-07
Num of attend

yongabyte:
The previous version of this branch would not do what it's set to do. Since the $scope.numOfattendant was only gonna keep track of one number, it won't properly display the correct number of attendance on a date if the date is not the most recent date.   

image


kmcnellis:
Thanks for catching that!

PR #501, opened 2016-10-18
Created a new panel on the projects page which shows the github commits for a project. For issue #457.

SaswatB:


client/app/projects/profile/profile.controller.js
```diff
@@ -18,6 +18,15 @@ angular.module('observatory3App')
                     $scope.checkUserProject();
                 }
             });
+            $('#github-commits').githubInfoWidget({ user: $scope.project.githubUsername, repo: $scope.project.githubProjectName, branch: 'master', last: 15 },
+              function() {
+                  console.log($('.github-user'));
```
agundy: Shouldn't commit the console.log

kmcnellis:
Nice! we should maybe think about styling it a little differently later (the darker inset looks a little off), but that might be part of an overhaul of the page.

screen shot 2016-10-25 at 12 34 40 pm

PR #506, opened 2016-10-24
Added some new tests

vkr96:


kmcnellis:
Could you list the new tests?

vkr96:
I only added two here- one actually creates a user session and the other test returns whether it is running ro nto

PR #509, opened 2016-10-27
Added seed files for achievements, added two new achievement tests

vkr96:
Added a test to check if uploading an achievement will work

Added a test to check if deleting an achievement will work

Added seed files for achievements


kmcnellis:
You should add a seed with a picture, and one with a multi-line description

PR #514, opened 2016-10-30
Fix bonus attendance recording

kmcnellis:
bonusDayCode was not defined, causing bonus and full day codes to not be distinguished

### DIFF
--- a/server/api/attendance/attendance.controller.js
+++ b/server/api/attendance/attendance.controller.js
@@ -102,7 +102,7 @@ exports.index = function(req, res) {
 exports.show = function(req, res) {
   Attendance.findById(req.params.id, function (err, attendance) {
     if(err) { return handleError(res, err); }
-    if(!attendance) { return res.send(404); }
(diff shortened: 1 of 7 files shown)


server/api/attendance/attendance.controller.js
```diff
@@ -185,15 +186,15 @@ exports.verifyFullAttendance = function(req,res){
     return getPresent(userId, isoDateToTime(new Date()), classYear._id, function(err,userAttendance){
       if (err) {return handleError(err)}
       if (!userAttendance.length){
-        return res.send(400, 'No attendance found');
+        return res.status(400).json('No attendance found');
```
agundy: this isn't valid json. How does it return?

kmcnellis: It returns properly. If you use .body it returns the string in the res.text field instead

agundy:
@kmcnellis to use send Json you need to give it a json object.

PR #517, opened 2016-10-31
Upgrade Deps

kmcnellis:
Remove some unused & upgrade others


agundy:
looks good except babel-runtime is needed for grunt serve:dist.

Issue #455, opened 2016-09-20
Github repo push integration


zmike808:

zmike808:
For now, I'm going to change the title from push integration to pull integration

zmike808:
PR submitted. @agundy could you look over it when you have some time?

agundy:
Thanks Mike I will!

PR #524, opened 2016-11-02
Feature/455 (Github pull integration #455)

zmike808:
Implemented feature #455. Only issue left is solving an async issue with the DB disconnecting before the commits finish saving. Besides that,  everything is working. So, I'd like to get my code reviewed and hopefully merged in.


kmcnellis:
Could you give a description of how everything works?

kmcnellis:
Could you give a d

server/workers/github.js
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+var token = config.GITHUB_WORKER_TOKEN;
+var octoclient = octo.client(token);
+
+Project.find().exec().then(
+  function (projects) {
+    projects.forEach(
+      function (project) {
+        console.error("PROJECT: ", project.fullRepoPath);
```
kmcnellis: Remove console.log

server/workers/github.js
```diff
@@ -0,0 +1,58 @@
...
+        var ghrepo = octoclient.repo(project.fullRepoPath);
+        ghrepo.commits(
+          function (err, commitDataArray, headers) {
+            commitDataArray.forEach(
+              function (commitData) {
+                var newCommit = new Commit(commitData);
+
+                // super hacky, definitely need to abstract this before github's api changes and this goes boom!!
```
kmcnellis: Is there any way to record the author or prevent loading the same commit twice?

server/workers/github.js
kmcnellis: What happens if the project url is invalid?

server/workers/github.js
```diff
@@ -0,0 +1,58 @@
...
+            );
+          }
+        );
+      }
+    );
+  }
+).then(
+  function () {
+    // @TODO FIX ASYNC ISSUES SO DISCONNECT HAPPENS AFTER ALL SAVES HAVE COMPLETED.
+    // Spent like 4 hours trying to debug this, but javascript is just too magical and asynchronous for me =(
```
kmcnellis: Comments like these are best for the issue or PR, so that people can discuss them. What have you tried?

kmcnellis:
How often would this be run? Is there any way for an admin to run a batch?

server/workers/github.js
```diff
@@ -0,0 +1,58 @@
...
+var octo = require('octonode');
+/*
+ @TODO: IMPORT TOKEN FROM AN UNTRACKED/PRIVATE CONFIG FILE
+ */
+
+var token = config.GITHUB_WORKER_TOKEN;
+var octoclient = octo.client(token);
+
+Project.find().exec().then(
+  function (projects) {
```
kmcnellis: It would be better if this was modular and could be run on only a specific set of projects

server/config/environment/development.js
```diff
@@ -17,5 +17,8 @@ module.exports = {
   // For testing verification
   // attendanceVerificationRatio: 1,
 
-  seedDB: false
+  seedDB: false,
+  // Github token for worker, local.env.js kept returning undefined but it worked once I added it here...
+  GITHUB_WORKER_TOKEN: 'YOUR_KEY',
```
kmcnellis: Check out server/config/environment/index.js for examples of how to read from the environment variables to set a config variable.

server/workers/github.js
```diff
@@ -0,0 +1,58 @@
...
+            commitDataArray.forEach(
+              function (commitData) {
+                var newCommit = new Commit(commitData);
+
+                // super hacky, definitely need to abstract this before github's api changes and this goes boom!!
+                newCommit.project = project;
+                newCommit.message = commitData.commit.message;
+                newCommit.date = commitData.commit.author.date;
+                newCommit.commentCount = commitData.commit.comment_count;
+                return newCommit.save();
```
kmcnellis: You will probably have to use promises. The problem is that db.disconnect happens  after the `projects.forEach loop is done, not waiting for `ghrepo.commits` or `newCommit.save`.

server/workers/github.js
```diff
@@ -0,0 +1,58 @@
...
+        );
+      }
+    );
+  }
+).then(
+  function () {
+    // @TODO FIX ASYNC ISSUES SO DISCONNECT HAPPENS AFTER ALL SAVES HAVE COMPLETED.
+    // Spent like 4 hours trying to debug this, but javascript is just too magical and asynchronous for me =(
+    // It works perfectly with db.disconnect commented out, since it will always disconnect before the saves occur...
+    // db.disconnect();
```
kmcnellis: You probably don't want to disconnect unless you connected in the `+if (!module.parent)` statement

server/workers/github.js
zmike808: The author field of the Commit model is currently populated & saved. As for preventing the loading of the same commit, do you mean when scraping Github, or actually running a save to the DB?

zmike808:
If you run "node ./server/workers/github.js" it will attempt to populate commits for all existing projects in the database. However, because the DB is not disconnecting, it will hang instead of closing after finishing. But you can browse the commits collection and view the data. Also, the commit api endpoints for users should be working. However, the project endpoint isn't working yet because the query is wrong.

server/workers/github.js
kmcnellis: 👍

agundy:
@zmike808 have a chance to make any of the changes that Kiana has suggested?

zmike808:
Made a ton of changes. Much cleaner code & everything is currently working. Please check it out and let me know what you guys think so far, @kmcnellis @agundy. ALSO: Make sure to run npm install, since I switched to a different github API wrapper/library

zmike808:
Status Update: almost done. Just need to integrate it with our endpoints. Know exactly what I have to do, just need to figure out how to do it in javascript & ExpressJS. Which sounds like it should be easy...but javascript. 

Anyways, I'm planning on making a new branch and PR once I have everything done. Just to make things cleaner.

agundy:
@zmike808 Sounds good! Let us know when we can look at it or if you need anything

kmcnellis:
I think its actually better to just leave it on this branch / pr, btw. It saves history for anyone looking back to see how it was done. It'll be squashed before merging anyways.

zmike808:
Hi, just wanted to drop a quick update. I know grades are due soon, and we were suppose to have everything in by today. I've been working on this last bit and it's been taking much longer than I thought. I am going to try to give it another hour or two, and if I can't get the api integration working, I'll just leave it as a function/on demand worker that can be run by nodejs cli.

zmike808:
I couldn't end up getting the endpoints to work. I ended up just adding a function into the worker which will grab commits for alll projects. I also added functions to the worker to get commits for a project with date filtering, i.e. before x, or after y.

PR #535, opened 2016-11-10
Project tests

vkr96:
Added two new tests that make sure a student can create a project and that an admin can delete a project. However, admins should not be able to delete project since that should only be allowed for mentors.


agundy:
Good start!

server/api/project/project.spec.js
```diff
@@ -54,4 +62,29 @@ describe('POST /api/projects/', function() {
...
+        websiteUrl: "www.google.com"
+      })
+      .end((err,res) => {
+        expect(res.status).to.equal(201);
+        done(err);
+     });
+  });
+
+  it('Should delete a project', done=> {
+    var test_id = "000000000000000000000010";
```
agundy: is this always going to be true? might be better to generalize this?

PR #539, opened 2016-11-16
Ansible Initialization and Deploy Scripts

agundy:
With this commit, a server can be initialized easily and then the
playbook to deploy the app can be run once all the dependencies are in
place. Assumes the local build in dist is what is desired on the host.
Also uses the local.env.js file. Later this can be set and copied over.
Once the initizalize script is run the `letsencrypt.sh` file must be run
once manually to go through the letsencrypt steps.


vkr96:
@agundy is there a way to add documentation on the README for how exactly to deploy?

agundy:
I think we should add it to the wiki or other docs. Readme is going to get
cluttered if we keep doing that

kmcnellis:
Can we add a link in the readme to a wiki article about it? Otherwise, looks fine. I don't really have the means to test it, though. Esp. without instructions in said wiki

PR #546, opened 2016-11-30
add checkings to prevent generating duplicate daycodes

yongabyte:
I am aware of the that the code's failing one of the code submitting test,
but I didn't make any change to the attendance file 
and I couldn't reproduce the same output manually on local host 
help! @agundy 
npm test: 
image

 local host:
user is jane
image

### DIFF
--- a/server/api/classyear/classyear.controller.js
+++ b/server/api/classyear/classyear.controller.js
@@ -6,6 +6,7 @@
 
 var ClassYear = require('./classyear.model');
 var Attendance = require('../attendance/attendance.model');
+var SmallGroup = require('../smallgroup/smallgroup.model');
 
 // Get current class year
 exports.index = function(req, res) {
@@ -151,7 +152,8 @@ exports.daycode = function(req, res){
     });
     return classYear.save(function(err, classYear){
       if (err) return handleError(res, err);
-      res.send(200, code);
+      return res.status(200).json(code);
+
     });
   });
 };
@@ -220,5 +222,20 @@ function generateCode(codeLength){
       var character = (Math.floor(Math.random() * characterOptions.length));
       code = code.concat(characterOptions[character.toString()]);
   }
+  ClassYear.findOne({"dayCodes.code":code})
+    .exec(function(err, classYear){
+      if (err) return handleError(res, err);
+      if(classYear) {
+        generateCode(codeLength);
+      }
+      else{
+        SmallGroup.findOne({"dayCodes.code":code})
+          .exec(function(err, smallgroup){
+            if (err) return handleError(res, err);
+            if(smallgroup){
+              generateCode(codeLength);
+            }
+        });}
+    });  
   return code;
 }
--- a/server/api/classyear/classyear.controller.js
+++ b/server/api/classyear/classyear.controller.js
@@ -142,18 +142,20 @@ exports.daycode = function(req, res){
         return res.send(200, classYear.dayCodes[i].code);
       }
     }
-    //Not ambigious code generator, function at the bottom.
-    var code = generateCode(6);
-
-    classYear.dayCodes.push({
-      date: today,
-      code: code,
-      bonusDay: req.body.bonusDay ? true : false
-    });
-    return classYear.save(function(err, classYear){
+    //unique code generator, function at the bottom.
+    uniqueDayCode(generateCode(6),6,function(err,dayCode){
       if (err) return handleError(res, err);
-      return res.status(200).json(code);
+      var code = dayCode;
 
+      classYear.dayCodes.push({
+        date: today,
+        code: code,
+        bonusDay: req.body.bonusDay ? true : false
+      });
+      return classYear.save(function(err, classYear){
+        if (err) return handleError(res, err);
+        return res.status(200).json(code);
+      });      
     });
   });
 };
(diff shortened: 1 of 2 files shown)


server/api/smallgroup/smallgroup.controller.js
```diff
@@ -304,5 +304,20 @@ function generateCode(codeLength){
...
+        generateCode(codeLength);
+      }
+      else{
+        SmallGroup.findOne({"dayCodes.code":code})
+          .exec(function(err, smallgroup){
+            if (err) return handleError(res, err);
+            if(smallgroup){
+              generateCode(codeLength);
+            }
+        });}
```
agundy: should probably bump the `}` to the next line to align with the else indentation

agundy:
The problem is the switch from `res.send(200, code)` to `res.code(200).json(code)` i'm pretty sure. will test

server/api/classyear/classyear.controller.js
```diff
@@ -220,5 +222,20 @@ function generateCode(codeLength){
       var character = (Math.floor(Math.random() * characterOptions.length));
       code = code.concat(characterOptions[character.toString()]);
   }
+  ClassYear.findOne({"dayCodes.code":code})
```
agundy: this code should not be inside of the generateCode function. generate code is supposed to only create the code. We should do the checks outside of the function. Or at least not attempt to send error messages inside this function. you have response sending in the function but the response object isn't passed into generateCode. This will cause an exception if there ever is a duplicate code.

yongabyte: @agundy if a duplicate code was found, it would call itself to get another code.

yongabyte:
@agundy when I changed it back to res.send(200, code), it failed 3 more attendance api tests

server/api/classyear/classyear.controller.js
yongabyte: @agundy the reason that I put the checking inside generateCode was meant to make sure when generateCode return a code to daycode, the returning code must not be a duplicate one.

server/api/classyear/classyear.controller.js
```diff
@@ -220,5 +222,20 @@ function generateCode(codeLength){
...
+    .exec(function(err, classYear){
+      if (err) return handleError(res, err);
+      if(classYear) {
+        generateCode(codeLength);
```
kmcnellis: Should be a return generateCode(codeLength);

server/api/classyear/classyear.controller.js
kmcnellis: This function is expected to be synchronous. By adding in the ClassYear.findOne, you are making it async. Anything within the findOne block can no longer correctly return a string to the calling function, and it will break. Also, by having a `handleError(res, err);` you are both trying to return an error and using variables that have not been defined in the function. Try hardcoding in a code that you know is in the database (for the top level of recursion only) to try out your code.  In the long term, you should probably write a async function that handles calling generateCode as many times as needed to get a unique one. I would also suggest that if you get a collision, increase the size of the code by 1 to avoid the possibility of failing many times in a row.

server/api/classyear/classyear.controller.js
```diff
@@ -220,5 +222,20 @@ function generateCode(codeLength){
...
+      }
+      else{
+        SmallGroup.findOne({"dayCodes.code":code})
+          .exec(function(err, smallgroup){
+            if (err) return handleError(res, err);
```
kmcnellis: This will cause an error because `res` and `err` are not defined.

server/api/classyear/classyear.controller.js
```diff
@@ -220,5 +222,20 @@ function generateCode(codeLength){
...
+            if(smallgroup){
+              generateCode(codeLength);
```
kmcnellis: If this function was a correct async one, you should have a `return generateCode(codeLength)`

server/api/smallgroup/smallgroup.controller.js
```diff
@@ -304,5 +304,20 @@ function generateCode(codeLength){
       var character = (Math.floor(Math.random() * characterOptions.length));
       code = code.concat(characterOptions[character.toString()]);
   }
+  ClassYear.findOne({"dayCodes.code":code})
```
kmcnellis: This entire segment doesn't currently get executed before the  return code at the bottom is finished. Therefore if it finds the code is duplicate, nothing but an exception would occur.

server/api/classyear/classyear.controller.js
```diff
@@ -141,17 +142,20 @@ exports.daycode = function(req, res){
...
-    var code = generateCode(6);
-
-    classYear.dayCodes.push({
-      date: today,
-      code: code,
-      bonusDay: req.body.bonusDay ? true : false
-    });
-    return classYear.save(function(err, classYear){
+    //unique code generator, function at the bottom.
+    uniqueDayCode(generateCode(6),6,function(err,dayCode){
```
yongabyte: for testing, you can pass in any existing code for the first parameter. e.g uniqueDayCode(any-code , 6 , ......

server/api/classyear/classyear.controller.js
```diff
@@ -219,6 +223,27 @@ function generateCode(codeLength){
...
   return code;
 }
+
+//Generating unique code.
+function uniqueDayCode(code,codeLength,callback){  
+  ClassYear.findOne({"dayCodes.code":code})
+    .exec(function(err, classYear){
+      if (err) return callback("error when getting dayCode",null);
+      if(classYear) {
+        return uniqueDayCode(generateCode(codeLength+1),codeLength+1,callback);
```
kmcnellis: This is ugly. Would be a lot cleaner if generateCode was just called at the beginning of uniqueDayCode. Then you don't need to define the code length twice.

kmcnellis:
Eventually we should move this to a common helper function, but this is a good solution for now!

kmcnellis:
@agundy ?

PR #549, opened 2016-12-06
add test for accessing user info

### LINKED ISSUE #213: Test that a getting another user doesn't have return their attendance
Attendance and grades are private data and shouldn't be returned from the server for anyone but the owner. For this please write a test that makes sure that the attendance field is absent from an unauthenticated call.

yongabyte:
fixes #213

### DIFF
--- a/server/api/user/user.spec.js
+++ b/server/api/user/user.spec.js
@@ -1,4 +1,6 @@
 'use strict';
+import {seed} from '../../config/seed';
+import {getSessionFor} from '../../auth/util.integration';
 
 var should = require('should');
 var app = require('../../app');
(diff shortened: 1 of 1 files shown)


server/api/user/user.spec.js
```diff
@@ -18,3 +20,35 @@ describe('GET /api/users/past', function() {
...
+  });
+
+  it('public profile should not contain private info ', done => {
+    var userId = "000000000000000000000400";
+
+    studentSession
+      .get('/api/users/' + userId)
+      .end((err, res) => {
+        if(res.body.attendance){
+          console.log("should not have access other/'s attudence")
```
agundy: Extra log line

vkr96:
why was this merged even though no checks passed @agundy ?

agundy:
Checks didn't pass because I merged it before it happened.

PR #550, opened 2016-12-06
test for accessing daycodes

### LINKED ISSUE #208: Test that only admins and mentors can get day codes
Attendence day codes need to be accessed by admins and mentors, make sure this works as expected. This may need a new test file that I can add.

yongabyte:
fixes #208 
add test for checking if only admin can get large group daycode
add test for checking if user can get dayCodes


agundy:
looks good. Merge after both pass

Issue #513, opened 2016-10-28
Attend page should show missed meetings


cowang4:
There is no way to see what days you missed attending. I'm thinking of adding this to the attend page. To do so I'll need to crosscheck the Attendance data with the SmallGroup and LargeGroup databases.

vkr96:
Okay, thanks for the help!

vkr96:
@cowang4 hows this going?

cowang4:
Hey. I've put aside some time this Sunday to work on this issue. I've taken a look at it and it seems like I just need to cross reference the SmallGroup and ClassYear dayCodes with the Attendance data in the getAttendance function [here](https://github.com/rcos/Observatory3/blob/master/server/api/attendance/attendance.controller.js#L250). It seems like the Attendance table only has the days that the user has submitted. And there is a SmallGroup model that has a [dayCodes](https://github.com/rcos/Observatory3/blob/239e5c6205414dc9f19538ecee0a66c7c7ab4336/server/api/smallgroup/smallgroup.model.js#L18) object that I can compare to. And the large group dayCodes are in the ClassYear table.
Admittedly I don't have much experience with Web development, so it's slow going.

PR #551, opened 2016-12-06
Attend Page shows Absences

cowang4:
Hey. So I wrote some code to show the user's missing attendances on the attend page. I'm new to this stuff so feedback is welcome. William made the absences show up red, but the UI is up for debate of course. Also, I had to change the select property to true on the attendance code for the smallgroup schema because the code is the only way to compare with the user's attendance objects. Is that a security vulnerability or something? I saw that it was just recently changed to not be included in the queries. Issue #513


agundy:
Hey looks great cowang4! I'll give feedback shortly! Sorry for the delay. Super appreciative of the work!

agundy:
That was a security issue where the current attendance code was being sent out and one could simply load the list of all codes and get the current day's code. This could be addressed differently if you need it for comparison. Maybe we should have another identifier that isn't the code, or just override the code if it is for the current day so the user can't see the most recent code.

cowang4:
Oh okay, I figured it might be something like that. Overriding it should be simple and easy. Or maybe we can remove all the codes from the objects before they get returned. But are there other functions that read from that database? Some other identifier would probably be cleaner in the long run.

kmcnellis:
I think it would best to do it by date, small group / large group & bonus/non-bonus.  That would allow students who switched small groups to not show a marked-absent for any days they attended in the old group.  Also, if you need the small group code, its best to leave the select:false and pull the codes just for your endpoint (see https://github.com/rcos/Observatory3/blob/master/server/api/attendance/attendance.controller.js#L371)

cowang4:
Okay, I changed the smallgroup comparison to be date and bonusDay based. Does that make more sense?

server/api/attendance/attendance.controller.js
```diff
@@ -251,7 +251,55 @@ var getAttendance = function(userId, classYearId, cb){
...
+          if (!found) {
+            attendance.push({date:smallattend.date, bonusDay:smallattend.bonusDay, smallgroup:true, verified:false, present:false});
+          }
+        }
+      }
+    });
+    ClassYear.find({current:true}, function (err, classyear) {
+      if (err) {return handleError(err)}
+      if (typeof(classyear) != 'undefined') {
+        for (var i = 0; i < classyear[0]["dayCodes"].length; i++) {
```
agundy: Instead of classyear[0] you should be able to use Mongooses `findOne` and a check for null instead of the check of typeof too

agundy:
I think the group and bonus day check is fine. We don't save the attendance object if it's the wrong code as far as I know.

server/api/attendance/attendance.controller.js
cowang4: Fixed

server/api/attendance/attendance.controller.js
```diff
@@ -251,7 +251,55 @@ var getAttendance = function(userId, classYearId, cb){
...
+          var smalldate = new Date(smallattend.date);
+          smalldate.setUTCHours(4); //sets time to midnight of EDT zone
+          smalldate.setUTCMinutes(0);
+          smalldate.setUTCSeconds(0);
+          smalldate.setUTCMilliseconds(0);
+
+          var found = false;
+          for (var j = 0; j < attendance.length; j++) {
+            var attend = attendance[j];
+            if (attend.date.getTime() == smalldate.getTime()) {
```
cowang4: Fixed

cowang4:
Okay. How about now?

agundy:
Looks good except for one tiny change. We are sorting by datetime and you don't initialize that variable. We can either change the sort to be just date or add a datetime parameter to the codes you are adding. I'm not sure why we have date and datetime as two unique fields. Kinda ugly but once you change that it looks great. Tested it and works well! Will merge after the fix, would have done it myself but i don't have push rights to your branch

agundy:
the date vs datetime difference is kinda ugly, not your design!

cowang4:
Yeah it displays date but sorts by datetime... pretty strange. I'll change it to sort by date.

Issue #537, opened 2016-11-15
Attendance Tests 


vkr96:
Create  a test that can create an attendance code and test other functionalities of the controller

vkr96:
Check out #552

PR #552, opened 2016-12-09
Added test that makes sure mentor can access attendacne json file

vkr96:
Adds an attendance test


kmcnellis:
You guys really have to get on admin `fixes #537` to the description so that the corresponding issue actually gets closed...

PR #557, opened 2016-12-14
Code cleanup

agundy:
Cleaned up a bunch of jshint errors on the server side. We still have work to do on the front end


yongabyte:
Don't forget to remove the getAttendees function on the frontend smallgroup controller. I removed it once, but it's still there.

PR #516, opened 2016-10-31
Attendance updates

kmcnellis:
Split admin into settings, users, and attendance 

Add a list of past attendance codes to admin, along with deleting codes and number of attendees.

Split code generation and showing past codes into a common directive for small and large group.

Smallgroups can create bonus daycodes

Students can no longer query the classyear or small group and receive a list of all attendance codes (just dates)

fixes #491

PR #561, opened 2017-01-26
Show smallgroup option to create if none

kmcnellis:
Missed this earlier


agundy:
Why are you removing the loaded? Shouldn't we wait until the response loads and if it's a 4XX set the loaded = true. If you remove the loaded it will show a create until the small group loads

kmcnellis:
Loaded was removed in #516 I'll see if it needs to be re-added instead

kmcnellis:
I'll leave re-adding to a future PR. This is just to rectify the feature not working.

agundy:
cool. Will merge

kmcnellis:
ping

PR #570, opened 2017-01-31
Deploy update

agundy:
Wahoo. Slowly improving the deploy flow. Still have a few manual steps for initial setup but now deploys will update with a single ansible run.


kmcnellis:
Thanks!

Issue #544, opened 2016-11-19
smallgroup tag on navbar not showing up for user


yongabyte:
on master

vkr96:
@yongabyte hows this going?

yongabyte:
I am super stuck at this one. I thought it just needed a simple api call to get user's smallgroup. 
but I couldn't find a perfect place to make this api call.

PR #579, opened 2017-02-07
show/hide smallgroup tag

yongabyte:
basic idea: show smallgroup tag when user is in a smallgroup hide otherwise.
a very late pr.
#544

### DIFF
--- a/client/components/auth/auth.service.js
+++ b/client/components/auth/auth.service.js
@@ -4,9 +4,13 @@ function AuthService($location, $http, $cookieStore, $q, appConfig, Util, User) 
     var currentUser = {};
     var safeCb = Util.safeCb;
     var currentUser = {};
+    var smallgroup = {};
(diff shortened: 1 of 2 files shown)


client/components/auth/auth.service.js
```diff
@@ -152,6 +156,7 @@ function AuthService($location, $http, $cookieStore, $q, appConfig, Util, User)
      */
     getCurrentUser(callback) {
       if (arguments.length === 0) {
+        smallgroup._id ? currentUser.smallgroup = true : currentUser.smallgroup = false;
```
agundy: I think a clearer way to write this might be `currentUser.smallgroup = smallgroup._id !== null;`

yongabyte: just found out that "null" actually wouldn't work for === or !== it has to be "undefined", interesting

agundy:
Great thanks @yongabyte

PR #609, opened 2017-02-18
Allow the target hosts to be set when running a playbook

agundy:


kmcnellis:
How do you set variable_host?

agundy:
Via th extra cars flag when calling from th cli

kmcnellis:
Can we get some documentation on this process in the wiki?

kmcnellis:
Still hoping for documentation, but merging

PR #613, opened 2017-02-28
Fixes #212

### LINKED ISSUE #212: Attempting to add a project while not authenticated should return a 401
Not sure what the exact code it throws is but I'm pretty sure it will throw a 403. Whatever the create a project endpoint throws to a non logged in user is the code we want to test that keeps being given.

zmike808:
Fixes #212

### DIFF
--- a/server/api/project/project.spec.js
+++ b/server/api/project/project.spec.js
@@ -10,26 +10,26 @@ var agent = request.agent(app);
 var project = request.agent();
 var test_id = "000000000000000000000010";
 
-describe('GET /api/projects', function() {
+describe('GET /api/projects', function () {
 
-  it('should respond with JSON array', function(done) {
+  it('should respond with JSON array', function (done) {
     request(app)
       .get('/api/projects')
       .expect(200)
       .expect('Content-Type', /json/)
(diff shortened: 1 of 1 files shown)


server/api/project/project.spec.js
```diff
@@ -10,129 +10,145 @@ var agent = request.agent(app);
...
+describe('GET /api/projects', function () {
 
-  it('should respond with JSON array', function(done) {
+  it('should respond with JSON array', function (done) {
     request(app)
       .get('/api/projects')
       .expect(200)
       .expect('Content-Type', /json/)
-      .end(function(err, res) {
+      .end(function (err, res) {
```
agundy: why the space here? We generally never put spaces here

zmike808: Think that webstorm did that when I did an auto format for the file. Is there a link to our coding style guidelines?

kmcnellis: https://github.com/rcos/Observatory3/blob/master/docs/CODINGSTANDARDS.md

kmcnellis: Looks like the space is fine

Issue #577, opened 2017-02-07
Add a favorite projects system


SaswatB:
Create a system that allows users to mark and track their favorite projects.

PR #614, opened 2017-03-02
Created a system to handle favorite projects for users

SaswatB:
This is for issues #577, #590, and #591.

### DIFF
--- a/client/app/projects/projects.controller.js
+++ b/client/app/projects/projects.controller.js
@@ -1,12 +1,33 @@
 'use strict';
 
 angular.module('observatory3App')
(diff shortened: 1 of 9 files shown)


connection-lost:
There is a few things that might interest you:
1. Make it possible to remove favorite via the sidebar. Consider you favorited a project and you want to un-favorite it, but the project is nowhere to be found because it is hidden in the "hidden projects".
2. Make the star change color / show a prompt while mouse is hovering on it. I see a lot of other sites doing this, which may make it more intuitive.

SaswatB:
I'll create a new issue for these features

Issue #605, opened 2017-02-14
Split blog feeds by semester


kmcnellis:
Should have a side menu that shows the semesters with blog posts. You click and it scrolls down or shows only those posts

Issue #368, opened 2016-03-21
Sorted Blog Entries 


vkr96:
Should sort blog entries based on semester and year. Currently, for a long standing project such as Observatory all blog posts are presented on the blog page, it would be nice if there could be folders based on the semester and school year. We should sort them so that the blog page is presented in a better way.

yongabyte:
closed by #616

PR #616, opened 2017-03-04
divide posts into semesters and a sidebar

yongabyte:
1. fix by adding custom comparator.
 <del>divide posts into semesters and use a object to wrap around the semesters 
 e.g  {fall 2016:[post1, post2..], spring 2016:[ ], .....}
 pitfall: the order of the semesters can't be sorted, because there are the properties of a object.</de>l

2. fix by add hidden-sm on the sidebar
 <del> when the browser's width reduces to a certain amount (e.g a portrait ipad), the side would get under the posts. I have made use of grid </del>.

3. Angularjs might not have native support for bootstrap scrollspy? 
<del> is there a body tag? how can I edit the body tag? I need it for scrollspy. 
[bootstrap scrollspy](http://v4-alpha.getbootstrap.com/components/scrollspy/) </del> 

 #605, #368

### DIFF
--- a/client/app/projects/blog/blog.controller.js
+++ b/client/app/projects/blog/blog.controller.js
@@ -1,7 +1,7 @@
 'use strict';
 
 angular.module('observatory3App')
-.controller('ProjectsBlogCtrl', function ($scope, $http, $stateParams, $uibModal, Auth, Project, notify) {
+.controller('ProjectsBlogCtrl', function ($scope, $http, $stateParams, $uibModal, $anchorScroll, $location, Auth, Project, notify) {
     $scope.isAuthor = false;
     $scope.isLoggedIn = Auth.isLoggedIn;
     $scope.isAdmin = Auth.isAdmin;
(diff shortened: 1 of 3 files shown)


agundy:
If you want to sort the dates you can get a list of the keys, sort that and then use that as a lookup to order the way you want, for now reverse chronological order seems fine though

yongabyte:
The semesters/keys list  does not the "reverse chronological order". it's pure insertion order. It might never cause any problems, but I just think someone should know.

client/app/projects/blog/blog.controller.js
```diff
@@ -80,12 +80,41 @@ angular.module('observatory3App')
...
+    }
+
+    var sortPosts = function(){
+      $scope.semesters = {};
+      for(var i =$scope.posts.length-1; i >-1 ; --i){
+        var postDate = $scope.posts[i].date;
+        var year = postDate.substring(0,4);
+        var semester = '';
+        if(postDate.charAt(5) > 0 || postDate.charAt(6) > 7)
+          semester = 'Fall';
```
SaswatB: braces on if and else

kmcnellis:
This fails
Error: [ngRepeat:dupes] Duplicates in a repeater are not allowed. Use 'track by' expression to specify unique keys. Repeater: sem in listOfsem, Duplicate key: string:Fall 2016, Duplicate value: Fall 2016

PR #623, opened 2017-03-14
Project Membership by Semester

jessamarie:
This PR intends to add functionality for the Project Members Project.

So far:
- Modified the Projects object in the user model to hold the project AND the semesters on the project
- Modified the api user controller:
  - removeProject removes entire project only if there is the current semester in the array OR just     
   removes the current semester if the user has multiple semesters on the project.
  - add project adds an entire project if it doesn't exist in the user's array OR just adds the current   
   semester on the project
- I have not yet done any client side modifications. 

Any tips on how to test this? Postman , maybe?

### DIFF
--- a/server/api/user/user.controller.js
+++ b/server/api/user/user.controller.js
@@ -12,6 +12,8 @@ var Commit = require('../commit/commit.model');
 var ClassYear = require('../classyear/classyear.model');
 var Attendance = require('../attendance/attendance.model');
 var SmallGroup = require('../smallgroup/smallgroup.model');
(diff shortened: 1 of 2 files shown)


kmcnellis:
I think the best way to test would be to write some server side api tests!

We'll also need to have a database migration script for the live server, and update the seed.

How about for now, we create a Feature branch (Feature-ProjectMembership), and merge all related pull requests into there? Then we get changes through PR faster, but still bring the full feature to master as one unit.

jessamarie:
Moved #627

PR #627, opened 2017-03-21
Project Membership by semester - API work

jessamarie:
This PR intends to add functionality for the Project Members Project.

So far:
- Modified the Projects object in the user model to hold the project AND the semesters on the project
- Modified the api user controller:
  - removeProject removes entire project only if there is the current semester in the array OR just     
   removes the current semester if the user has multiple semesters on the project.
  - add project adds an entire project if it doesn't exist in the user's array OR just adds the current   
   semester on the project
- I have not yet done any client side modifications. 

Any tips on how to test this? Postman , maybe?

Same as #623

### DIFF
--- a/server/config/seed/users.json
+++ b/server/config/seed/users.json
@@ -2,6 +2,12 @@
     "provider": "local",
     "name": "Test User",
     "email": "test@test.com",
+    "projects": [{
+      "project": "000000000000000000000010",
+       "classYears": [{"id":"000000000000000000000001",
+                        "joinDate": "this is totally a date string"}],
+       
+    }],
     "password": "test",
     "github": {
         "login": "test"
--- a/server/api/user/user.controller.js
+++ b/server/api/user/user.controller.js
@@ -625,17 +625,59 @@ exports.resetPassword = function(req, res){
 exports.addProject = function(req,res){
     var userId = req.params.id;
     var newProject = req.body.project;
+
     User.findById(userId, function(err,user){
-        if (err){
-            res.send(500, err);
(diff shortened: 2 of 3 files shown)


server/config/seed/users.json
```diff
@@ -2,6 +2,12 @@
     "provider": "local",
     "name": "Test User",
     "email": "test@test.com",
+    "projects": [{
+      "project": "000000000000000000000010",
+       "classYears": [{"id":"000000000000000000000001",
+                        "joinDate": "this is totally a date string"}],
```
kmcnellis: You should replace this with an actual date!

server/config/seed/users.json
```diff
@@ -2,6 +2,12 @@
     "provider": "local",
     "name": "Test User",
     "email": "test@test.com",
+    "projects": [{
+      "project": "000000000000000000000010",
+       "classYears": [{"id":"000000000000000000000001",
```
kmcnellis: Lets keep this consistent, and change classYears.id into classYears. classYear

kmcnellis: Add a newline after the {

server/config/seed/users.json
kmcnellis: Add a newline before the }

jessamarie: Will do!

jessamarie:
I will try to make these changes and also start working on the migration script by Tuesday's small group session!

kmcnellis:
Merging into the feature branch!

PR #632, opened 2017-03-24
Fix showing blog posts

kmcnellis:
Fixes #616


yongabyte:
Isn't the duplicates of the same semester situation caused by mistakenly pushing same semester into the list? Since you fixed it, the list no longer has duplicates in it. Why would track by $index be necessary?

kmcnellis:
Its a generally good thing to include in ng-repeats

Issue #617, opened 2017-03-04
Bug on adding the project


bobmayuze:
![image](https://cloud.githubusercontent.com/assets/14257524/23581736/25e57e90-00e8-11e7-986f-42e02183de83.png)

When I was adding my project, I put "https://github.com/bobmayuze" in the first block and "Borilliant_Feedback" in the second block. It works, but I can not access my project anyway.

jfucci:
Any additional info? this [link](https://rcos.io/projects/bobmayuze/borilliant_feedback/profile) seems to let me access your project just fine.

Issue #604, opened 2017-02-14
Multi-repositories drop down not working


kmcnellis:
Issues:
The main repository is replaced by a blank link, linking back to the home page
The button spacing is off

kmcnellis:
See https://rcos.io/projects/smaltby/canary/profile

Issue #603, opened 2017-02-14
Link to project proposal should not be required


agundy:
When saving a project you can't save without having a link to a proposal. This means that any old projects can not be edited without adding a link. This is an annoying requirement that shouldn't be enforced for user flow. Especially because a new project may not have a proposal link right away

zmike808:
Fixed in PR #635

PR #635, opened 2017-03-30
Input validation on the project creation form

jfucci:
githubUsername and githubProjectName having slashes in them was causing issues. This code makes it so that they cannot have a forward slash.

should fix issue #617

### DIFF
--- a/client/components/editProject/editProject.html
+++ b/client/components/editProject/editProject.html
@@ -27,18 +27,19 @@
         <p class="help-block" ng-show="editProjectForm.githubUsername.$error.required && submitted">
           The project owner's github username is required
         </p>
-        <p class="help-block" ng-show="editProjectForm.githubUsername.$error.mongoose && submitted">
-          This field cannot have a '/' character
-        </p>
       </div>
       <div class="form-group" ng-class="{ 'has-success': editProjectForm.githubProjectName.$valid && submitted,
       'has-error': editProjectForm.githubProjectName.$invalid && submitted }">
         <p class="help-block" ng-show="editProjectForm.githubProjectName.$error.required && submitted">
           The project's github name is required
         </p>
-        <p class="help-block" ng-show="editProjectForm.githubProjectName.$error.mongoose && submitted">
-          This field cannot have a '/' character
-        </p>
+      </div>
+      <div class="form-group" ng-class="{
+          'has-success': editProjectForm.githubUsername.$valid && editProjectForm.githubProjectName.$valid && submitted,
+          'has-error': (editProjectForm.githubUsername.$invalid || editProjectForm.githubProjectName.$invalid) && submitted }">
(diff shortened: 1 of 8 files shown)


jfucci:
should fix issue #617

jfucci:
newest commit fixes #603... which also has to do with input validation

jfucci:
latest 2 commits fix issue #604

kmcnellis:
Once the field is verified with a / character, the warning never goes away & the user cannot submit
screen shot 2017-04-11 at 1 37 04 pm

kmcnellis:
The various inputs with verification should have live error-checking using 
`ng-model-options="{ updateOn: 'default blur', debounce: { 'default': 500, 'blur': 0 } }" `
as seen in https://github.com/rcos/Observatory3/blob/master/client/app/account/settings/settings.html

client/components/editProject/editProject.html
```diff
@@ -27,12 +27,18 @@ <h3 ng-hide="editing" class="modal-title">New Project</h3>
...
+        <p class="help-block" ng-show="editProjectForm.githubUsername.$error.mongoose && submitted">
+          This field cannot have a '/' character
+        </p>
       </div>
       <div class="form-group" ng-class="{ 'has-success': editProjectForm.githubProjectName.$valid && submitted,
       'has-error': editProjectForm.githubProjectName.$invalid && submitted }">
         <p class="help-block" ng-show="editProjectForm.githubProjectName.$error.required && submitted">
           The project's github name is required
         </p>
+        <p class="help-block" ng-show="editProjectForm.githubProjectName.$error.mongoose && submitted">
```
kmcnellis: These two `This field cannot have a '/' character` messages shouldn't stack. Lets just use 
`ng-show="(editProjectForm.githubUsername.$error.mongoose || editProjectForm.githubProjectName.$error.mongoose) && submitted"` or something of the sort

package.json
```diff
@@ -80,7 +80,6 @@
     "grunt-newer": "^1.1.1",
     "grunt-ng-annotate": "^1.0.1",
     "grunt-ng-constant": "^2.0.1",
-    "grunt-node-inspector": "^0.4.1",
```
kmcnellis: This shouldn't really be in this PR, but ok. 🤷‍♂️

jfucci: It was the first thing I had to do to get grunt serve to run so I just committed it. Can revert if necessary

jfucci:
I cannot reproduce the issue with not being able to re-send the form after sending one with slashes. Are you sure the branch is fully up to date? I had that issue at some point, but not now:
http://imgur.com/a/7N7YZ

client/components/editProject/editProject.html
jfucci: latest commit fixes this

package.json
kmcnellis: Yeah, it doesn't work with the newest version of node. You can leave it in

kmcnellis:
Looks good now, thanks.

PR #638, opened 2017-04-02
Fix typo on login form

copperwater:
The login form says "Need and Account?". This changes it to "an".


connection-lost:
This fix is awesome. I can see a massive increase to the website's user-friendliness. Great job.

PR #644, opened 2017-04-12
Tag system

yongabyte:
1. remove duplicate tag code on user and project profile.
2. create tag directive for all tag.
3. tag for blog.
4. change tag bubble color from #eee to #ddeeff.
 #eee is the same color as the background color for blog post. 
image

feature of tag for blog:
1.tags will only get save when user trigger submit post.
2. remove the "+" button  -------------      on the editblog page for tag. 
                                        | new tag |  +   <-----removed
                                        --------------
Whenever + is hit, submit will be trigger, user forced to exit blog edit popup. Problem did not occur when hitting enter for the same function.
blogs:
image
edit blog:
image

Honorable mention: @SaswatB

### DIFF
--- a/client/app/projects/blog/blog.controller.js
+++ b/client/app/projects/blog/blog.controller.js
@@ -13,8 +13,10 @@ angular.module('observatory3App')
     };
 
     $scope.userOwnsPost = function(post) {
-        var userId = $scope.user._id;
-        return ($scope.isAdmin() || $scope.getCurrentUser().role.toLowerCase() === 'mentor' || userId === post.author._id);
+        if($scope.user._id){
+          var userId = $scope.user._id;
+          return ($scope.isAdmin() || $scope.getCurrentUser().role.toLowerCase() === 'mentor' || userId === post.author._id);
+        }
     };
 
     $scope.editPost = function(post) {
@@ -143,7 +145,9 @@ angular.module('observatory3App')
 
       });
     };
-
+    $scope.readOnly = function(){
+      return false;
+    }
     $scope.load();
 
   });
--- a/client/app/projects/blog/blog.less
+++ b/client/app/projects/blog/blog.less
@@ -6,3 +6,7 @@
     height: 150px;
     width: 100%;
 }
+.blog-tag .tech-tags {
+	padding : 0px;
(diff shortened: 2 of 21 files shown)


client/app/projects/blog/blog.less
```diff
@@ -6,3 +6,7 @@
     height: 150px;
     width: 100%;
 }
+.blog-tag .tech-tags {
+	padding : 0px;
+	margin-top: -10px;
```
SaswatB: Spacing does not seem correct

server/api/user/index.js
```diff
@@ -36,10 +36,10 @@ router.put('/:id/deactivate', auth.canEdit(), controller.deactivate);
 router.put('/:id/activate', auth.isAuthenticated(), controller.activate);
 router.put('/:id/project', auth.isAuthenticated(), controller.addProject);
 router.put('/:id/favorite/:project', auth.isAuthenticated(), controller.addFavorite);
-router.put('/:id/addTech', auth.canEdit(), controller.addTech);
-router.put('/:id/removeTech', auth.canEdit(), controller.removeTech);
+router.put('/addTechBubble/:id/:tech', auth.canEdit(), controller.addTech);
+router.put('/:id/:tech/removeTech', auth.canEdit(), controller.removeTech);
 
-router.delete('/:id/project', auth.isAuthenticated(), controller.removeProject);
+router.delete('/addTechBubble/:id/:tech', auth.isAuthenticated(), controller.removeProject);
```
SaswatB: please follow convention of /:id/ first

client/app/projects/blog/blog.controller.js
```diff
@@ -143,7 +145,9 @@ angular.module('observatory3App')
 
       });
     };
-
+    $scope.readOnly = function(){
```
SaswatB: Please use consistent naming.
Below this is called canEdit(), unless there's a specific reason to call this readOnly() it would be ideal to use the same name.

kmcnellis:
@SaswatB ?

yongabyte:
@agundy what need to be fixed?

kmcnellis:
@agundy ?

PR #654, opened 2017-04-29
Set the max age to 1 year for quicker page loads

agundy:


agundy:
Merging since this got no response and it's a small fix

PR #660, opened 2017-09-13
Fix past user page ids

agundy:
Few linting fixes and switched the past users endpoint to have string and not binary ids

### DIFF
--- a/server/api/user/user.model.js
+++ b/server/api/user/user.model.js
@@ -247,7 +247,7 @@ UserSchema
   .virtual('listInfo')
   .get(function() {
     return {
-      '_id':this._id.toString('binary'),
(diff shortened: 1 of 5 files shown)


server/api/user/user.model.js
```diff
@@ -247,7 +247,7 @@ UserSchema
   .virtual('listInfo')
   .get(function() {
     return {
-      '_id':this._id.toString('binary'),
+      '_id':this._id.toString(),
```
yongabyte: Nice found. I applied the same changes to line 221, and line 203  it solved the errors of unreable id when getting profiles. Shall we apply the changes to all of those lines?

agundy: sounds great!

PR #702, opened 2018-03-07
Commit Controller Documentation (Issue #693)

ra0s:


aeksco:
There are a few problems here!

First, remove the `docs/` folder from the repository - we shouldn't be checking in the generated API docs, mainly because they'll be changing so often that tracking those changes will bloat the repository. It also looks like you deleted the `docs/` directory - but there wasn't really anything in there of particular relevance at this juncture so no worries :)

Second - I think you might need to do `git pull upstream dev` - I think you're missing some changes that were previously merged in.

The only file that really should have changes is the Commit Controller - touch base with me on slack if you hit any snags!

ra0s:
I removed the folder and turns out I did delete the other folder by accident. I ran `git pull upstream dev` and I think everything should be lined up now

PR #734, opened 2018-04-10
Code Quality for achievement.controller.js

trane2:
Changed 'var' to 'const' and replaced 'function' with new syntax '=>'


andrew-domino:
wrong trane... you mean @trane2 :)

aeksco:
@trane2 I'll actually merge this one in since there's nothing here's that's causing problems - just make sure you open another PR with your changes to the `classyear.controller.js` file - thanks!

PR #753, opened 2018-04-16
Updated user controller actions to return JSON with HTTP Status

### LINKED ISSUE #713: User controller actions must return JSON with HTTP status
Encapsulates issues:  #706 | #707 | #708 | #709 | #712 

The User module's controller (`/server/api/user/user.controller.js`) methods return only HTTP status codes (i.e. `200`, `301`, `404`, `500`). These responses should changed to return, at the very least, a simple JSON representation of what has occurred. 

The following statements in `user.controller.js` should be updated:

Before:
```
res.send(500, err);

res.send(200);

res.send(403);

```

After:
```
res.status(500).json({ error: err })

res.status(200).json({ success: true })

res.status(403).json({ forbidden: true })
```

We will likely update these statements to return the _data_ that has changed as a result of the API call - this issue will pave the way for those changes.

AndrewMwakibinga:
Resolved #713


aeksco:
Closing - reopen against `dev`

Issue #725, opened 2018-04-07
Abstract handleError function in Achievement.controller.js


aeksco:
Import the `handleError` function from `helpers.js` into `achievement.controller.js` and delete the `handleError` function defined there.
 
Replace the following statements to ensure proper functionality.
Replace this:
```
    .catch(handleError(res));
```

With this:
```
    .catch((err) => handleError(res, err));
```

aeksco:
Resolved

PR #757, opened 2018-04-17
Finish inport handleError into achievement.controller.js, and upgrade…

YZT0119:
#725 Finish import and code upgrading


aeksco:
:+1: great work!

Issue #772, opened 2018-09-30
UrpForm model


ratchet12340:
Here is the specification for an UrpForm model. For context on what the functionality will be here, we will have an entry of an UrpForm for each URP form that each student fills out. This way, we will have a big listing of the all the URP forms, and Prof. Goldschmidt can sign off on the whole batch with one click.

- UrpForm
  - user_id (ref)
  - semester_id (ref)
  - upload_url (string, references filepath)

The role of this issue is to create the needed database model so that Mongo will know what properties an 'UrpForm' document should have (basically, just the 3 properties we have above). 

The model for posts may be a good reference (found in `server/api/post/post.model.js`).

PR #778, opened 2018-10-02
Did the URP Model

dchau2017:
See title


aeksco:
:+1: Nice - good first pass. 

@dchau2017 Make sure you reference the issue in your PR comment, like so: #772

PR #790, opened 2018-10-16
Added team model

### LINKED ISSUE #787: Team model
Here is the specification for an Team model. For context on what the functionality will be here, we will have an entry for each team for each semester. 

This way, we can keep track of what the current teams are and who is working on which project each semester.
Team
 - title (string)
 - user_ids ([ref])
 - semester_id (ref)
 - project_id (ref)

The role of this issue is to create the needed database model so that Mongo will know what properties a 'Team' document should have (basically, just the 4 properties we have above).

The model for posts may be a good reference (found in server/api/post/post.model.js).

dchau2017:
Resolved #787 
I added the team model.

### DIFF
--- a/server/api/team/team.model.js
+++ b/server/api/team/team.model.js
@@ -1,0 +1,37 @@
+'use strict';
+
+var mongoose = require('mongoose'),
+    Schema = mongoose.Schema;
+
+var TeamSchema = new Schema({
+  title: String,
+  content: String,
+
+  title:
+  {
+    type: String,
+    index: true
+  },
+
+
+  userId: 
+  {
+    type : Schema.Types.ObjectId,
+    ref: 'User',
+    index: true
+  },
+  semesterId: 
+  {
+    type : Schema.Types.ObjectId,
+    ref: 'ClassYear',
+    index: true
+  },
+  projectId: 
+  {
+    type : Schema.Types.ObjectId,
+    ref: 'Project',
+    index: true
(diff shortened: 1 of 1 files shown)


server/api/team/team.model.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+var mongoose = require('mongoose'),
+    Schema = mongoose.Schema;
+
+var TeamSchema = new Schema({
+  title: String,
```
aeksco: Delete this - duplication at line`10`

server/api/team/team.model.js
```diff
@@ -0,0 +1,37 @@
...
+  content: String,
```
aeksco: Renamed `content` attribute to `description`

server/api/team/team.model.js
```diff
@@ -0,0 +1,37 @@
...
+
+  title:
+  {
+    type: String,
+    index: true
+  },
+
+
+  userId:
```
aeksco: This should be an array an objects, like so: 
```
  userId: [{
    type : Schema.Types.ObjectId,
    ref: 'User',
    index: true
  }],
```

aeksco (changes requested):
Looks great - just adjust to my comments and we'll get it merged in :)

aeksco (approved):
Great work, thanks for fixing those issues!

Issue #789, opened 2018-10-16
Controller for Team


TungE:
Implement the necessary function that each API route uses in a controller file for the Team API.

aeksco:
Resolved

Issue #788, opened 2018-10-16
API routes for Team model


TungE:
The role of this issue is to create the needed API routes so that we can do things like get all teams by semester. Here is a list of routes we should have with the associated RESTful verb:

    Get ALL Teams
    Get ALL Teams by semester
    Get a specific Team by project & semester
    Update a Team
    Delete a Team

aeksco:
Resolved

PR #791, opened 2018-10-16
added barebones index and controller file for Team Model

TungE:
This addresses the following two issues: #788  #789 
Also added the route to the routes.js


aeksco:
:+1:

aeksco (approved):
:+1: nice work!

Issue #796, opened 2018-10-23
Update ClassYear.getCurrent() function and associated calls throughout app


aeksco:
The `ClassYear.getCurrent()` uses a callback - it should return a "then-able" object so it can be used with `async/await` and Promises. See below:

```js
// in classyear.model.js
ClassSchema.statics.getCurrent = function(cb){
  return ClassYear.findOne({ current: true })
};
```

Update all the locations this function is called in the application - do a find-all for `ClassYear.getCurrent`.

Also update the `ClassYear.getCurrentCodes` function to return a "then-able" as well.

aeksco:
This can be implemented in the controllers like so:

```js
const classYear = await ClassYear.getCurrent().catch((err) => handleError(err))
```

Love that `async/await` with the `catch` statement :+1:

Issue #792, opened 2018-10-23
Replace callbacks with async/await in smallgroup.controller.js 


aeksco:
All our Express.js controllers use callbacks when interacting with our Mongoose models - like so:

```js
exports.index = (req, res) => {
    SmallGroup.find({}, (err, smallgroups) => {
        if (err) { return handleError(res, err); }
        return res.status(200).json(smallgroups);
    });
};
```
These functions should be updated to use the [async/await](https://medium.com/tech-buddy/async-await-without-try-catch-in-javascript-fdd38abf7e90) syntax, like so:

```js
exports.index = async (req, res) => {
  const smallgroups = await SmallGroup.find({}).catch((err) => { return handleError(res, err); })
  return res.status(200).json(smallgroups);
};
```

@TungE LMK if you've got any questions - sorry for the delay!

aeksco:
@TungE Just a note - start doing this in the `smallgroup.controller.js` file - work on one function at a time, and commit your changes after you update each function. You can run `npm test` from the `observatory-server` root to ensure that your changes haven't broken anything :+1:

PR #797, opened 2018-10-23
refactored all of room.controller and most of smallgroup.controller

TungE:
Mostly addresses #792

- manually fixed spacing/indentation in both files
- refactored all of room.controller to use try-catch and async/await (instead of callbacks and .then().catch())
- refactored most of smallgroup.controller in the same way BUT two functions remain untouched; these two functions must be refactored in my process of resolving issue #796 

"npm test" ran with no errors, although I'm new this code, so there will probably be some hidden errors and other general things I missed

### DIFF
--- a/server/api/room/room.controller.js
+++ b/server/api/room/room.controller.js
@@ -11,14 +11,13 @@ import Room from './room.model'
 * @apiSuccess {Collection} root Collection of all the user's Rooms.
 * @apiError (500) UnknownException Could not retrieve Room collection
 */
-exports.index = (req, res) => {
-  return Room.find()
-  .then((collection) => {
+exports.index = async (req, res) => {
+  try {
+    const collection = await Room.find()
     return res.json(200, collection).end()
-  })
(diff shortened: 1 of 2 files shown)


server/api/room/room.controller.js
```diff
@@ -11,14 +11,13 @@ import Room from './room.model'
 * @apiSuccess {Collection} root Collection of all the user's Rooms.
 * @apiError (500) UnknownException Could not retrieve Room collection
 */
-exports.index = (req, res) => {
-  return Room.find()
-  .then((collection) => {
+exports.index = async (req, res) => {
```
aeksco, line 14: Looks great! You could slim this down a bit like so:

```js
exports.index = async (req, res) => {
    const collection = await Room.find().catch((err) => return handleError(res, err))
    return res.status(200).json(collection)
}
```

aeksco (approved):
Looks great! Take a look at the comment I left - nice pattern to remove those `try/catch` statements

PR #798, opened 2018-10-27
Added routes for User Registration

dchau2017:
I added routes for User Registration

### DIFF
--- a/server/api/user_registration/index.js
+++ b/server/api/user_registration/index.js
@@ -1,0 +1,18 @@
+'use strict';
+
+const express = require('express');
+const controller = require('./project.controller');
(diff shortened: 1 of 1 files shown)


server/api/user_registration/index.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+const express = require('express');
+const controller = require('./project.controller');
```
aeksco: This should import something like:

```js
const controller = require('./user_registration.controller');
```

Just fix that issue and we'll merge it in!

aeksco (commented):
Address feedback please and thanks!

Issue #808, opened 2018-11-18
Replace callbacks with async/await in user.controller.js 


TungE:
All our Express.js controllers use callbacks when interacting with our Mongoose models.

These functions should be updated to use the async/await syntax.

See #792 for reference.

PR #809, opened 2018-11-18
refactored user controller

TungE:
Addresses #808.

Most functions (as appropriate) were refactored to use the async/await syntax (rather than callbacks), more consistent whitespacing, simplified logic flow, and updated a few instances of deprecated functions (res.json(status, obj) --> res.status(status).json(obj)).

Notably, the publicStats and destroy functions have not been refactored. Alex S. mentioned that the destroy function has some blatant security flaw and should be looked at.

In using "npm test", all tests passed except for 1 pending, which is pending because expiresInMinutes and expiresInSeconds are deprecated, although I'm not sure how I can fix this on my end. This may have stopped the rest of the user controller from even being tested.


aeksco:
:+1: Terrific work - merging this in :shipit:

PR #810, opened 2018-11-20
replaced deprecated expiresInMinutes and refactored user controller c…

TungE:
…reate

Also removed required field for first_name and last_name in user model


aeksco:
:+1: looks great - will merge after passing testing

Issue #805, opened 2018-11-06
Excused Absence - deny and update


off-by-1-error:
1) The `exports.deny` controller function needs to be defined
2) In `exports.update`, the behavior needs to change depending on the status of the user

aeksco:
Resolved by #811

PR #811, opened 2018-11-20
Resolved Issue #805

off-by-1-error:
Added deny and update implementations, also rewrote the exports.destroy function in the controller


aeksco (approved):
:+1: Terrific work!

Issue #802, opened 2018-11-05
Mentors/Coordinators should not need to verify themselves 


vakrao:
If someone is a mentor/coordinator they should not need to be verfied

PR #814, opened 2018-11-27
now, mentors should not have to verify themselves for attendance

### LINKED ISSUE #815: Replace deprecated express functions identified by the tests ("npm test")
Change uses of res.json(status, obj) to res.status(status).json(obj)
Change uses of res.status(status) to res.sendStatus(status)

TungE:
I haven't tested this change, but I added a simple check in the attendance controller which should make it so that mentors do not need to verify themselves for attendance. Should resolve issue #802 .

Edit: now also resolved #815 .


aeksco (approved):
Terrific work - merging!

